PR #17, opened 2019-06-14
Add ratio for all as another type of treatment application

wenzeslaus:
I can merge if there are no objections

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project should be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## 2019-06-14
+
+### Added
+
+- Added types of treatment application. Removing a portion for both
+  susceptible and infected, like in SLF case, and a portion from
+  susceptible all all infected in given cells, like in SOD case.
+  (Vaclav Petras)
+
 ## 2018-11-17
 
 - Added mortality function to simulation class
--- a/TECHNICALDEBT.md
+++ b/TECHNICALDEBT.md
@@ -17,6 +17,11 @@ issue which needs to be fixed (Fix).
 Entries should be removed when resolved. Issue from tracker can be
 optionally linked in an entry.
 
+## 2019-06-14
+
+### Add
+
+- Documentation to previously implemented treatments is missing
 
 ## 2018-09-14 - SEID and multiple host handling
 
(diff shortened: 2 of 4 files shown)


dagaydos21:
Hi Vashek, I'm just curious what you mean by this. Do you mean that we would be removing a ratio of all the species in a treatment area?

wenzeslaus:
@dagaydos21 The current treatment code removes all infected individuals in the cell (as if the treatment raster is only 0s and 1s) and then, if treatment raster values are in between 0-1, removes percentage of the susceptible individuals. This works well for SOD and cases where you always want to remove 100%. The new code adds an option to remove ratio from both I and S. This new option is the default as you can always pass in a raster of 0s and 1s and 1s will translate into 100% removal.

Suggestions for better naming are more than welcome!

...and yes, the above should be in the doc and is not. I added note to technical debt file.

dagaydos21:
Okay, two questions: 1) what happens when you have 1 infection but have a ratio of removal, let's say .45. Does it remove the 1 infection or leave it?  and 2) Is this an option I can specify, or do I need to convert all of the management raster data I currently have?

wenzeslaus:
> 1) what happens when you have 1 infection but have a ratio of removal, let's say .45. Does it remove the 1 infection or leave it?

It will remove it. We are using the default double-to-integer conversion which is dropping the fractional part. We can do something different, but given the expression (the calculation we use), it gives expected result in most cases. I would say that 1 and ratios around 0.5 are edge cases. Here is how this plays out in the code (both computed by hand and actually tested with code):

```
int(1 - 1 * 0.45) = int(1 - 0.45) = int(0.55) = 0
int(10 - 10 * 0.45) = int(10 - 4.5) = int(5.5) = 5
int(100 - 100 * 0.45) = int(10 - 45) = int(55) = 55
```

...again, perhaps something for doc and even tests.

> 2) Is this an option I can specify, or do I need to convert all of the management raster data I currently have?

No need. Both the R package and GRASS module (once they have this), will provide two options, one will work as it worked before (called `AllInfectedInCell` in C++).

dagaydos21:
Okay thanks, this helps clear things up. It'll be interesting to see how this works with the SOD data. There are a lot of cases where there's only 1 infection, so it might essentially just work the same. I'd say go ahead and merge it if Chris doesn't have any objections.

ChrisJones687:
Thanks Vashek,

This is exactly what I had in mind. Please merge it when you have time and
I will update the R version so this is a selectable option.  Thanks again!

Best,
Chris

wenzeslaus:
Thanks. Sounds good. I'm working on the GRASS module now.

Just for the record: About the integer computations, on top of what I said before, integer-double conversions might be something to revisit for future releases.

PR #27, opened 2019-08-12
Complete rewrite of dispersal kernels

### LINKED ISSUE #26: resolution should be double not int
https://github.com/ncsu-landscape-dynamics/PoPS/blob/master/simulation.hpp#L128

or am I missing something?

wenzeslaus:
Each kernel or category of kernels is in its own class
including composite kernels which are switching between
different kernels based on parameters or random numbers.

The results are (slighly) different due to change from
wrong representation of resolution as int (now double)
and distance (double) using now the abs(double) function
instead of its integer (and thus inappropriate) version.

Fixes #26.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,38 @@ All notable changes to this project should be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## 2019-08-11 - Dispersal kernel rewrite
+
+### Added
+
+- Radial exponential kernel and a uniform kernel (Vaclav Petras)
+  * Any combination of kernels is now possible on the library level.
+- Functions to convert from strings to kernel-related enums.
+
+### Changed
+
+- Dispersal kernel are organized as classes (functors). (Vaclav Petras)
+  * The disperse function is now a function template in the Simulation.
+  * Dispersal kernel is a function (functor) called from the Simulation.
+  * The Simulation class does not have any kernel-related parameters
+    or members.
+- Usage of std::cout related to kernels was replaced by exceptions
+  in the new code. (Vaclav Petras)
+- Kernel-related enums are not strongly typed using enum class
+  (Vaclav Petras)
+
+### Fixed
+
+- A call to abs function on the distance used for spread caused
+  conversion to int. Now std::abs is used so double is used throughout.
+  This changes simulation results. (Vaclav Petras)
+- Resolution is now double instead of int. This changes results in cases
+  when the resolution is not a integer number (even when it is
+  so close to the closest higher integer that it is printed out without
+  any decimal places by software considering it, possibly correctly,
+  close enough from user perspective). This was also reported in the
+  issue #26. (Vaclav Petras)
+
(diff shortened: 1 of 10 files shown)


wenzeslaus:
I have tested this with the GRASS module and the results are completely the same without the resolution and abs() fixes and they are almost the same with the fixes. The resolution fix changes results only in special, yet quite real, cases (e.g. resolution=99.999...999). See below the univariate statistics of difference raster of probabilities (n=4) between version with and without abs() fix (sanity check: spatial results "look" the same).

```
n: 471705
minimum: -100
maximum: 100
range: 200
mean: 0.386735
mean of absolute values: 5.98483
standard deviation: 14.5438
variance: 211.523
variation coefficient: 3760.66 %
sum: 182425
```

Documentation for usage is missing, but doc for adding kernels is there.

Here is the code form the GRASS module. The idea is to set parameters for each kernel separately and then nest them. The nesting happens in a predefined way, i.e. how that can be done is defined in the code. Enabling and disabling in runtime is handled by the parameters as opposed to using dynamic polymorphism which would be the alternative here. (A test comparing syntax and speed to dynamic polymorphism would be of course good to have, but I don't have it at this point.)

```
RadialDispersalKernel short_radial_kernel(window.ew_res, window.ns_res,
                                          rtype, scale1, pwdir, kappa);
RadialDispersalKernel long_radial_kernel(window.ew_res, window.ns_res,
                                         rtype, scale2, pwdir, kappa);
UniformDispersalKernel uniform_kernel(window.rows, window.cols);
SwitchDispersalKernel short_selectable_kernel(rtype, short_radial_kernel, uniform_kernel);
SwitchDispersalKernel long_selectable_kernel(rtype, long_radial_kernel, uniform_kernel);
DispersalKernel kernel(short_selectable_kernel,
                       long_selectable_kernel,
                       use_long_kernel,
                       gamma);

sporulations[run].disperse(sus_species_rasts[run],
                           inf_species_rasts[run],
                           inf_species_cohort_rasts[run][current_age],
                           lvtree_rast,
                           outside_spores[run],
                           weather,
                           weather_coefficients[step],
                           kernel);
```

A simple example using just RadialDispersalKernel directly in Simulation::disperse() from test is:

```
RadialDispersalKernel kernel(ew_res, ns_res, dispersal_kernel,
                             short_distance_scale);
simulation.disperse(susceptible, infected,
                    mortality_tracker, total_plants,
                    outside_dispersers, weather, weather_coefficient,
                    kernel);
```

ChrisJones687:
Tested with R version and updated all of the functions in rpops to use the new dispersal kernel classes.

wenzeslaus:
Thanks. I will merge ncsu-landscape-dynamics/r.pops.spread#11 which I just updated to create kernels just once at the beginning of the simulation. There is no difference in results with the current kernels comparing to creating kernels for each step (at least with GCC on Linux, C++11 distributions may have a state in general, see e.g. [std::cauchy_distribution::reset()](http://www.cplusplus.com/reference/random/cauchy_distribution/reset/)).

PR #37, opened 2019-10-31
Fix raster and scalar operators

wenzeslaus:
Plus and minus (r+s) were broken from 87eda93d. The overloading was ambiguous
after 804f21fae. Minus and divide (s-r) were broken copy-pasted in 87eda93d.

### DIFF
--- a/raster.hpp
+++ b/raster.hpp
@@ -281,46 +281,6 @@ public:
     }
 
     template<typename OtherNumber>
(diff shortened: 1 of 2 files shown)


wenzeslaus:
@petrasovaa please review things in the code such as operator plus actually doing plus internally and that tests are testing what they are claiming to test (such as operators and test reference data matching).

Issue #34, opened 2019-10-07
Add Resistant pool(s)


ChrisJones687:
Add in the resistant pool(s). The idea here is that it is another compartment in the model but one that is ignored for most calculations of spread but is used more for treatment using pesticides (related to #33 ) or if resistant is obtained when an individual recovers from a disease (unaware of this in plant systems but happens often in human and animal diseases).

This will include a function to return hosts from the resistant pool to the susceptible pool after a defined length of time. The question is whether option 1) one large pool that could be imported and exported with a bunch of small trackers with separate return times option or 2) have a bunch of separate pools that are linked to their return times that can be imported/exported makes the most sense and is the most flexible. 

Started something here in [a rpops branch](https://github.com/ncsu-landscape-dynamics/rpops/tree/feature/fungicide_treatment). We just need to think through the best way to implement this in an easy to use and flexible manner.

PR #40, opened 2019-12-16
added current_resistant to apply_treatment()

ChrisJones687:
added current resistant to the apply treatment function in order to account for temporal overlap. Previous if there was a temporal overlap of pesticide treatments all current resistant pools were overwritten with 0 thus causing pesticide treatment to be treated more like culling or host removal.

### DIFF
--- a/test_treatments.cpp
+++ b/test_treatments.cpp
@@ -218,6 +218,69 @@ int test_combination()
     return num_errors;
 }
 
+
+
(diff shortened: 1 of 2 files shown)


ChrisJones687:
@petrasovaa there was an issue with time but not space overlap the former resistant pool where new treatment wasn't applied was reset to 0.

ChrisJones687:
fixes #34

PR #41, opened 2019-12-18
WIP: scheduling system for simulation (spread, treatments, export, mortality, lethal temp)

petrasovaa:
This is a work-in-progress system to simplify and standardize scheduling events in simulation. A simple example is in tests:

    Date st(2020, 1, 1);
    Date end(2021, 12, 31);

    // define simulation time step, in this case every 2 months
    Scheduler scheduling(st, end, StepUnit::Month, 2);
    // get vector of bools if spread should happen in step N
    std::vector<bool> spread_schedule = scheduling.schedule_spread(Season(1, 12));

    // scheduling regular events:
    // do certain action (e.g. export) at the end of the year
    std::vector<bool> export_schedule =  scheduling.schedule_action_yearly();
    // do certain action (e.g. export) every N time steps (in this case 2* 2 months)
    std::vector<bool> export_schedule =  scheduling.schedule_action_nsteps(2);

    // scheduling irregular events (treatments)
    unsigned n = scheduling.schedule_action_date(Date(2020, 3, 3));

    // in the simulation loop
    for (unsigned i = 0; i < scheduling.get_num_steps(), ++i) {
        if (spread_schedule[i])
            // simulate spread
        if (export_schedule[i])
            // export results
        ...
    }

    // example of schedule with seasonality
    Date st(2020, 2, 1);
    Date end(2021, 12, 18);
    Scheduler scheduling (st, end, StepUnit::Month, 3);
    std::vector<bool> vect = scheduling.schedule_spread(Season(3, 9));
    std::cout << vect << std::endl;

    2020-2-1 - 2020-4-30: true
    2020-5-1 - 2020-7-31: true
    2020-8-1 - 2020-10-31: true
    2020-11-1 - 2021-1-31: false
    2021-2-1 - 2021-4-30: true
    2021-5-1 - 2021-7-31: true
    2021-8-1 - 2021-10-31: true
    2021-11-1 - 2022-1-31: false

Not everything is done yet (e.g. integrating this with treatments, documentation). The advantage is, it's in the library, so it would be standardized across rpops/r.pops.spread. Let me know what you think about this approach.

### DIFF
--- a/date.hpp
+++ b/date.hpp
@@ -162,11 +162,11 @@ bool Date::is_last_day_of_year()
 bool Date::is_last_week_of_month()
 {
   if (this->is_leap_year()){
-    if ((day_ + 7) == day_in_month[1][month_])
+    if ((day_ + 7) >= day_in_month[1][month_])
       return true;
     return false;
   } else {
-    if ((day_ + 7)== day_in_month[0][month_])
+    if ((day_ + 7) >= day_in_month[0][month_])
       return true;
     return false;
   }
--- a/date.hpp
+++ b/date.hpp
@@ -1,7 +1,7 @@
 /*
  * SOD model - date manipulation
  *
- * Copyright (C) 2015-2017 by the authors.
+ * Copyright (C) 2015-2019 by the authors.
  *
  * Authors: Zexi Chen (zchen22 ncsu edu)
  *          Anna Petrasova
@@ -59,6 +59,8 @@ public:
     inline bool is_last_week_of_year();
     inline bool is_last_month_of_year();
     inline bool is_last_day_of_year();
+    inline bool is_last_day_of_month();
+    inline bool is_last_week_of_month();
     inline bool is_leap_year();
     int month() const { return month_; }
     int year() const { return year_; }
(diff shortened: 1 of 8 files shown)


ChrisJones687:
I really like this concept. It moves many of the checks we have for exporting and simulating into a single workflow and simplifies the loop which should minimize the chance for difference across wrappers. 

I added is_last_day_of_week() and is_last_day_of_month() to date.hpp in order to do monthly exporting if running daily or weekly time step but exporting monthly

We would still need this for daily time step with a weekly output (i.e. is_last_day_of_week() function) currently implemented in rpops as just export every 7 steps within the simulation (if it is daily with weekly output) but there is definitely a better way to do this.

date.hpp
```diff
@@ -157,6 +159,32 @@ bool Date::is_last_day_of_year()
     return false;
 }
 
+bool Date::is_last_week_of_month()
```
petrasovaa: This is not going to work, this implementation expects 4 weeks in a month. It needs to test for certain interval of days.

petrasovaa:
I don't think we need these functions for the new scheduling implementation. The scheduling class uses intervals as steps, which I think simplifies it conceptually. I will add functions to schedule actions monthly, add more documentation.

date.hpp
ChrisJones687: Good catch it should have been >= not ==.

PR #43, opened 2019-12-18
Avoid creating a raster in simulation

wenzeslaus:
Now caller needs to pass an existing dispersers raster to the simulation.
The con that it needs to exist and user needs to keep track of it,
but the pro is that it can be examined and used further or even modified
in between steps of simulation.

Consequently the dispersers are no longer a hidden detail, but part of what is
exposed to the caller.

The motivation is the avoiding of raster creation in simulation which makes
the simulation less dependent on particular raster class behavior,
in this case, a specific constructor (int, int) is no longer needed.

### DIFF
--- a/simulation.hpp
+++ b/simulation.hpp
@@ -143,7 +143,7 @@ public:
      * `std::make_tuple(row, column)` fulfills this requirement.
      */
     template<typename DispersalKernel>
-    void disperse(IntegerRaster& dispersers,
+    void disperse(const IntegerRaster& dispersers,
                   IntegerRaster& susceptible, IntegerRaster& infected,
                   IntegerRaster& mortality_tracker,
                   const IntegerRaster& total_plants,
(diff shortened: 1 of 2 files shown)


wenzeslaus:
This might be advantageous for wrappers such as the pybind11 one, although at this point I can work without it. My main question is whether this creates unnecessary burder for rpops or r.pops.spread?

All what is needed in r.pops.spread is something like

```
IntegerRaster dispersers(infected.rows(), infected.cols());
```

right before `Simulation::generate()` and `Simulation::spread()` calls and passing `dispersers` as a first parameter of these two functions. I assume that rpops will need something similar.

I'm not constrained by this, but we need to resolve it before 1.0.0.

ChrisJones687:
I really like this for 2 reasons:
1. dispersers are now exportable as well so we can see which location is producing the most spores at a given output.
2. it would allow for creating a total_dispersers (i.e. the total number of spores generated for the each location for the entire simulation) which could be very useful for highlighting these locations along with other types of important locations in the interfaces as a way to provide more information for managers to use.

It shouldn't cause unnecessary difficulties for rpops.

PR #46, opened 2020-02-04
Feature/movement

ChrisJones687:
added the movement module to the simulation class. This is currently working and implemented in the rpops package. I made a few changes from what we had done based on integer overflow issues.

### DIFF
--- a/simulation.hpp
+++ b/simulation.hpp
@@ -164,26 +164,33 @@ public:
             int row_to = moved[2];
             int col_to = moved[3];
             int hosts = moved[4];
-            if (infected(row_from, col_from) > 0 || susceptible(row_from, col_from) > 0) {
-                inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
-                int infected_mean = infected(row_from, col_from) * inf_ratio;
-                std::poisson_distribution<int> distribution(infected_mean);
-                if (infected_mean > 0)
-                    infected_moved += distribution(generator_);
-            }
             if (hosts > total_plants(row_from, col_from)) {
-              total_hosts_moved = total_plants(row_from, col_from);
+                total_hosts_moved = total_plants(row_from, col_from);
             } else {
-              total_hosts_moved = hosts;
+                total_hosts_moved = hosts;
             }
-            if (infected_moved > infected(row_from, col_from)) {
-              infected_moved = infected(row_from, col_from);
-            } 
-            if ((total_hosts_moved - infected_moved) > susceptible(row_from, col_from)) {
-              susceptible_moved = susceptible(row_from, col_from);
+            if (infected(row_from, col_from) > 0 && susceptible(row_from, col_from) > 0) {
+                inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
+                int infected_mean = total_hosts_moved * inf_ratio;
+                if (infected_mean > 0) {
+                    std::poisson_distribution<int> distribution(infected_mean);
+                    infected_moved += distribution(generator_);
+                }
+                susceptible_moved = total_hosts_moved - infected_moved;
+                if (infected_moved > infected(row_from, col_from)) {
+                    infected_moved = infected(row_from, col_from);
+                } 
+                if (susceptible_moved > susceptible(row_from, col_from)) {
+                    susceptible_moved = susceptible(row_from, col_from);
+                } 
+            } else if (infected(row_from, col_from) > 0 && susceptible(row_from, col_from) == 0) {
+                infected_moved = total_hosts_moved;
+            } else if(infected(row_from, col_from) == 0 && susceptible(row_from, col_from) > 0) {
+                susceptible_moved = total_hosts_moved;
             } else {
-              susceptible_moved = total_hosts_moved - infected_moved;
+                continue;
             }
+            
             infected(row_from, col_from) -= infected_moved;
             susceptible(row_from, col_from) -= susceptible_moved;
             total_plants(row_from, col_from) -= total_hosts_moved;
--- a/simulation.hpp
+++ b/simulation.hpp
@@ -164,7 +164,7 @@ public:
             int row_to = moved[2];
             int col_to = moved[3];
             int hosts = moved[4];
-            if (infected(row_from, col_from) > 0 || susceptible(row_from, col_from) > 0) {
+            if (infected(row_from, col_from) > 0 && susceptible(row_from, col_from) > 0) {
                 inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
                 int infected_mean = infected(row_from, col_from) * inf_ratio;
                 std::poisson_distribution<int> distribution(infected_mean);
@@ -186,11 +186,6 @@ public:
             }
             infected(row_from, col_from) -= infected_moved;
             susceptible(row_from, col_from) -= susceptible_moved;
-            if (hosts > total_plants(row_from, col_from)) {
-                total_hosts_moved = total_plants(row_from, col_from);
-            } else {
-                total_hosts_moved = hosts;
-            }
             total_plants(row_from, col_from) -= total_hosts_moved;
             infected(row_to, col_to) += infected_moved;
             susceptible(row_to, col_to) += susceptible_moved;
--- a/simulation.hpp
+++ b/simulation.hpp
@@ -157,7 +157,6 @@ public:
             int infected_moved = 0;
             int susceptible_moved = 0;
             int total_hosts_moved = 0;
-            double infected_mean = 0;
             double inf_ratio = 0;
             int row_from = moved[0];
             int col_from = moved[1];
@@ -179,6 +178,9 @@ public:
                 if (infected_moved > infected(row_from, col_from)) {
                     infected_moved = infected(row_from, col_from);
                 } 
+                if (infected_moved > total_hosts_moved) {
+                  infected_moved = total_hosts_moved;
+                }
                 susceptible_moved = total_hosts_moved - infected_moved;
                 if (susceptible_moved > susceptible(row_from, col_from)) {
                     susceptible_moved = susceptible(row_from, col_from);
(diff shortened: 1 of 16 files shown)


simulation.hpp
```diff
@@ -129,6 +129,75 @@ class Simulation
...
+                inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
+                infected_mean = infected(row_from, col_from) * inf_ratio;
+                std::poisson_distribution<int> distribution(infected_mean);
+                if (infected_mean > 0)
+                    infected_moved += distribution(generator_);
+                }
+            if ((hosts - infected_moved) > susceptible(row_from, col_from)) {
+                susceptible_moved = susceptible(row_from, col_from);
+            } else {
+                susceptible_moved = hosts - infected_moved;
```
wenzeslaus: infected_moved (result of poisson) can be huge, so susceptible_moved can become negative and later `infected(row_to, col_to)` too large unless I'm missing some condition here. It seems like an additional case is missing to account for this state which is basically "all moved hosts are the infected ones." The infected_moved needs to be limited by infected(row_from, col_from). If I'm thinking about this correctly, this theoretically increases probability of having all moved host infected.

Perhaps you could also put all checks for data input first (like num_hosts <= total_host, num_hosts <= susceptible() + infected()) and then resolve the other issues as they appear resulting in correct move counts. Perhaps this way it would be easier to see if we account for all the special cases.

This is a clear candidate for precondition/postcondition/assert for total_plants == infected + susceptible, but I'm not too excited about adding these to the code when the only tool for it is the assert macro. However, this is something what a separate test in test_simulation.cpp could check for.

simulation.hpp
```diff
@@ -129,6 +129,75 @@ class Simulation
...
+            }
+            infected(row_from, col_from) -= infected_moved;
+            susceptible(row_from, col_from) -= susceptible_moved;
+            if (hosts > total_plants(row_from, col_from)) {
+                total_hosts_moved = total_plants(row_from, col_from);
+            } else {
+                total_hosts_moved = hosts;
+            }
+            total_plants(row_from, col_from) -= total_hosts_moved;
+            if (infected(row_from, col_from) < 0)
```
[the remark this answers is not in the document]
ChrisJones687: removed

simulation.hpp
```diff
@@ -129,6 +129,75 @@ class Simulation
...
+            double infected_mean = 0;
+            double inf_ratio = 0;
+            int row_from = moved[0];
+            int col_from = moved[1];
+            int row_to = moved[2];
+            int col_to = moved[3];
+            int hosts = moved[4];
+            if (infected(row_from, col_from) > 0 || susceptible(row_from, col_from)) {
+                inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
+                infected_mean = infected(row_from, col_from) * inf_ratio;
```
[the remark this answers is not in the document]
ChrisJones687: good call!

simulation.hpp
```diff
@@ -129,6 +129,75 @@ class Simulation
...
+            int susceptible_moved = 0;
+            int total_hosts_moved = 0;
+            double infected_mean = 0;
+            double inf_ratio = 0;
+            int row_from = moved[0];
+            int col_from = moved[1];
+            int row_to = moved[2];
+            int col_to = moved[3];
+            int hosts = moved[4];
+            if (infected(row_from, col_from) > 0 || susceptible(row_from, col_from)) {
```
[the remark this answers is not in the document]
ChrisJones687: yes good catch

simulation.hpp
ChrisJones687: I adjusted this piece to first check hosts moved aren't greater than total hosts and then do the infected and susceptible checks based on the output from the previous check and to ensure that the infected draw from the Poisson distribution wasn't larger than the infected in the cell.

simulation.hpp
```diff
@@ -129,6 +129,72 @@ class Simulation
...
+            int col_to = moved[3];
+            int hosts = moved[4];
+            if (infected(row_from, col_from) > 0 || susceptible(row_from, col_from) > 0) {
+                inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
+                int infected_mean = infected(row_from, col_from) * inf_ratio;
+                std::poisson_distribution<int> distribution(infected_mean);
+                if (infected_mean > 0)
+                    infected_moved += distribution(generator_);
+            }
+            if (hosts > total_plants(row_from, col_from)) {
```
wenzeslaus: This needs to be done first as infected_moved should be computed from total_hosts_moved.

simulation.hpp
```diff
@@ -129,6 +129,72 @@ class Simulation
...
+            int total_hosts_moved = 0;
+            double infected_mean = 0;
+            double inf_ratio = 0;
+            int row_from = moved[0];
+            int col_from = moved[1];
+            int row_to = moved[2];
+            int col_to = moved[3];
+            int hosts = moved[4];
+            if (infected(row_from, col_from) > 0 || susceptible(row_from, col_from) > 0) {
+                inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
```
petrasovaa: if S=0, then we have division by zero

simulation.hpp
```diff
@@ -129,6 +129,72 @@ class Simulation
...
+                int infected_mean = infected(row_from, col_from) * inf_ratio;
```
petrasovaa: we think, the ratio should be multiplied by hosts, not infected.

simulation.hpp
```diff
@@ -129,6 +129,72 @@ class Simulation
...
+            }
+            if (hosts > total_plants(row_from, col_from)) {
+              total_hosts_moved = total_plants(row_from, col_from);
+            } else {
+              total_hosts_moved = hosts;
+            }
+            if (infected_moved > infected(row_from, col_from)) {
+              infected_moved = infected(row_from, col_from);
+            } 
+            if ((total_hosts_moved - infected_moved) > susceptible(row_from, col_from)) {
```
wenzeslaus: Compute susceptible_moved first, then test it.

simulation.hpp
```diff
@@ -129,6 +129,72 @@ class Simulation
...
+            double inf_ratio = 0;
+            int row_from = moved[0];
+            int col_from = moved[1];
+            int row_to = moved[2];
+            int col_to = moved[3];
+            int hosts = moved[4];
+            if (infected(row_from, col_from) > 0 || susceptible(row_from, col_from) > 0) {
+                inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
+                int infected_mean = infected(row_from, col_from) * inf_ratio;
+                std::poisson_distribution<int> distribution(infected_mean);
```
petrasovaa: If the goal of using the distribution is to randomize how many I vs S will be picked from the total number of moved, then I suggest applying the distribution on the inf_ratio, maybe gaussian then? That would make it more clear I think and we would probably avoid a lot of the subsequent checks.

ChrisJones687: The issue with using a gaussian is we would need to specify a standard deviation parameter. With a Poisson distribution, we don't need to specify a standard deviation or any other parameter other than the expected mean. It is why they are used frequently in disease models.

simulation.hpp
```diff
@@ -129,6 +129,79 @@ class Simulation
...
+            int row_to = moved[2];
+            int col_to = moved[3];
+            int hosts = moved[4];
+            if (hosts > total_plants(row_from, col_from)) {
+                total_hosts_moved = total_plants(row_from, col_from);
+            } else {
+                total_hosts_moved = hosts;
+            }
+            if (infected(row_from, col_from) > 0 && susceptible(row_from, col_from) > 0) {
+                inf_ratio = double(infected(row_from, col_from)) / double(susceptible(row_from, col_from));
```
petrasovaa: This should actually be

    inf_ratio = double(infected(row_from, col_from)) / double(total_plants(row_from, col_from));

This will BTW get more complex with the exposed.

simulation.hpp
```diff
@@ -129,6 +129,79 @@ class Simulation
...
+                int infected_mean = total_hosts_moved * inf_ratio;
+                if (infected_mean > 0) {
+                    std::poisson_distribution<int> distribution(infected_mean);
+                    infected_moved += distribution(generator_);
```
petrasovaa: Minor, but it should be = not +=. Using += looks like it will be added several times which is not the case here.

simulation.hpp
```diff
@@ -129,6 +129,79 @@ class Simulation
...
+                }
+                susceptible_moved = total_hosts_moved - infected_moved;
```
petrasovaa: The susceptible_moved should be shifted below so that it's computed after we trim the infected:

    if (infected_moved > infected(row_from, col_from)) {
        infected_moved = infected(row_from, col_from);
    } 
    susceptible_moved = total_hosts_moved - infected_moved;
    if (susceptible_moved > susceptible(row_from, col_from)) {
        susceptible_moved = susceptible(row_from, col_from);
    }

simulation.hpp
```diff
@@ -129,6 +129,79 @@ class Simulation
...
+                inf_ratio = double(infected(row_from, col_from)) / double(total_plants(row_from, col_from));
+                int infected_mean = total_hosts_moved * inf_ratio;
+                if (infected_mean > 0) {
+                    std::poisson_distribution<int> distribution(infected_mean);
+                    infected_moved = distribution(generator_);
+                }
+                if (infected_moved > infected(row_from, col_from)) {
+                    infected_moved = infected(row_from, col_from);
+                } 
+                susceptible_moved = total_hosts_moved - infected_moved;
```
petrasovaa: Looking at it again, I think there should be one more check:

    if (infected_moved > total_hosts_moved)
        infected_moved = total_hosts_moved

ChrisJones687: this should be taken care of with the previous check as we are assuming that infected + susceptible = total_plants and thus infected can't be larger than total hosts. That said it is probably better to be explicit here and this check shouldn't increase computational time.

petrasovaa: If inf_ratio is 1, then infected_moved can be easily > total_hosts_moved.

wenzeslaus: I agree that the check is needed because even if we limit by the total available infected that does not tell anything about how many hosts we actually plan to move. We can have 50 infected hosts, but we might be moving only 20 total (so at most 20 infected).

ChrisJones687:
Closes #48

PR #52, opened 2020-03-24
Helper functions for parsing input + scheduler fix

petrasovaa:
Contains helper functions needed for both rpops and r.pops.spread to parse treatment application enum, step unit enum, and function returning output schedule that unifies handling of combinations of simulation step and output step.

Also includes fix for scheduler for cases when given end date of simulation == start date of a last simulation step, now this last simulation step is included in the simulation (e.g. weekly simulation from 1/1 to 1/8 would result in 2 steps - 1/1-1/7 and 1/8-1/14, before this only results in 1/1-1/7).
This is consistent with the previous behavior, if the end date of last interval was after end date of the simulation, the last interval is included.

### DIFF
--- a/scheduling.hpp
+++ b/scheduling.hpp
@@ -20,6 +20,9 @@
 
 #include <iostream>
 #include <vector>
+#include <map>
+#include <tuple>
+#include <string>
 #include <algorithm>
 
 #include "date.hpp"
@@ -58,6 +61,30 @@ std::ostream& operator<<(std::ostream& os, const Step &step)
 enum class StepUnit {
     Day, Week, Month
 };
+
+/**
+ * @brief Get step enum from string
+
+ * Throws an std::invalid_argument exception if the value
+ * is not supported.
+ */
+inline StepUnit step_unit_enum_from_string(const std::string& text)
+{
+    std::map<std::string, StepUnit> mapping{
+        {"day", StepUnit::Day},
+        {"week", StepUnit::Week},
+        {"month", StepUnit::Month}
+    };
+    try {
+        return mapping.at(text);
+    }
+    catch (const std::out_of_range&) {
+        throw std::invalid_argument("step_unit_enum_from_string:"
+                                    " Invalid value '" + text +"' provided");
+    }
+}
+
+
 
 class Scheduler
 {
@@ -114,6 +141,13 @@ public:
      */
     Step get_step(unsigned index) const {
         return steps.at(index);
+    }
+
+    /**
+     * @brief Get length of simulation step as number of units and unit type
+     */
+    std::tuple<unsigned, StepUnit> get_step_length() const {
+        return std::make_tuple(simulation_num_units_, simulation_unit_);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


scheduling.hpp
```diff
@@ -59,13 +62,39 @@ enum class StepUnit {
...
 class Scheduler
 {
 public:
     /**
      * Scheduler creates a vector of simulation steps
      * based on start, end date, unit and number of units.
      * The steps can be e.g. 1 day, 3 months, 2 weeks.
+     * Step is scheduled if start date of an interval is
+     * within [start, end] even if end date of the step is outside.
      * @param start simulation start date
```
wenzeslaus: Usually empty line is used before list of parameters.

scheduling.hpp
```diff
@@ -59,13 +62,39 @@ enum class StepUnit {
...
+
 class Scheduler
 {
 public:
     /**
      * Scheduler creates a vector of simulation steps
      * based on start, end date, unit and number of units.
      * The steps can be e.g. 1 day, 3 months, 2 weeks.
+     * Step is scheduled if start date of an interval is
+     * within [start, end] even if end date of the step is outside.
```
wenzeslaus: This describes behavior, but please add why it is useful, i.e., "...so that the whole requested interval is contained in the resulting steps..." or sth like that.

test_scheduling.cpp
```diff
@@ -283,6 +284,98 @@ int test_get_number_of_scheduled_actions()
...
+}
+
+int test_output_schedule_from_string()
+{
+    int num_errors = 0;
+    Date st(2020, 1, 1);
+    Date end(2020, 1, 28);
+
+    Scheduler scheduling(st, end, StepUnit::Week, 1);
+    std::vector<bool> out = output_schedule_from_string(scheduling, "week", 0);
```
wenzeslaus: Not having the zero here explicitly is probably better since the function already supports that. Nevertheless, one test of this case (unused but explicitly provided) should be included since that might be how the function will be used.

treatments.hpp
```diff
@@ -36,6 +37,27 @@ enum class TreatmentApplication {
...
+    std::map<std::string, TreatmentApplication> mapping{
+        {"ratio_to_all", TreatmentApplication::Ratio},
+        {"all_infected_in_cell", TreatmentApplication::AllInfectedInCell}
+    };
+    try {
+        return mapping.at(text);
+    }
+    catch (const std::out_of_range&) {
+        throw std::invalid_argument("treatment_application_enum_from_string:"
+                                    " Invalid value '" + text +"' provided");
```
wenzeslaus: Missing space in: text + "'

treatments.hpp
```diff
@@ -36,6 +37,27 @@ enum class TreatmentApplication {
...
+/**
+ * @brief Get treatment application enum from string
+
+ * Throws an std::invalid_argument exception if the value
+ * is not supported.
+ */
+inline TreatmentApplication treatment_app_enum_from_string(const std::string& text)
+{
+    std::map<std::string, TreatmentApplication> mapping{
+        {"ratio_to_all", TreatmentApplication::Ratio},
```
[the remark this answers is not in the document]
wenzeslaus: I support having alternative versions in the library. I already decided for that in the SI/SEI code: https://github.com/ncsu-landscape-dynamics/PoPS/pull/51/files#diff-f728f36c8a8c4becfad6d804458a3849R56

Each interface can decide what makes sense in its context and the lib can support that by accepting more than one version. This is especially true in case of different spellings (dashes, underscores, case, ...).

petrasovaa:
I consider the StepUnit as unit, not frequency, so 'daily' does not make much sense in the interface I designed it with - the simulation step/frequency is given by unit and number of units so that we can run the simulation every 2 weeks for example. I added 'daily' and others to the output frequency.

PR #57, opened 2020-05-15
Add operator to assign a single scalar to raster

wenzeslaus:
This makes initialization easier when only one value is needed
and it does what contructors are doing.
However, it makes sense only for Raster which won't change
into a scalar (unlike other similar classes), so it can be used
for tests, but not within PoPS itself.

### DIFF
--- a/raster.hpp
+++ b/raster.hpp
@@ -281,6 +281,13 @@ public:
     }
 
     template<typename OtherNumber>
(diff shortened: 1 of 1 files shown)


wenzeslaus:
Unlike the corresponding constructor, this shortcut for fill may not be expected behavior. Closing as won't fix for now.

PR #58, opened 2020-05-15
Optionally remove stochasticity in disperser generation and establishment

wenzeslaus:
Stochasticity can be now reduced by turning it off in generate function and in the
establishment step of the disperse function.

### DIFF
--- a/simulation.hpp
+++ b/simulation.hpp
@@ -113,6 +113,8 @@ class Simulation
 private:
     RasterIndex rows_;
     RasterIndex cols_;
+    bool dispersers_stochasticity_;
+    bool establishment_stochasticity_;
     ModelType model_type_;
     unsigned latency_period_;
     std::default_random_engine generator_;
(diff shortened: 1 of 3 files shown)


simulation.hpp
```diff
@@ -342,7 +355,8 @@ class Simulation
                   std::vector<std::tuple<int, int>>& outside_dispersers,
                   bool weather,
                   const FloatRaster& weather_coefficient,
-                  DispersalKernel& dispersal_kernel)
+                  DispersalKernel& dispersal_kernel,
+                  double establishment_probability = 1)
```
ChrisJones687, line 359: I think the default should be 0.5 as it would allow for some establishment, otherwise, everything >0 will establish. This is good for general testing of the functionality via tests but not ideal for testing the effect of the random effects on uncertainty outcomes (this is our most likely use case).

wenzeslaus: Makes sense. Done and documented.

Issue #50, opened 2020-02-18
Double check parameter names to make sure they are general enough


ChrisJones687:
A good example is total_plants to total_hosts

Issue #30, opened 2019-08-14
Rename ShortLongDispersalKernel to natural/anthro or to sth more generic


wenzeslaus:
The ShortLongDispersalKernel class template follows the original naming convention of short  and long distance dispersal. Now we call it natural and anthropogenic. The kernel class itself is more general, so there is more options and it is just a question of naming.

```
ShortLongDispersalKernel(short_kernel, long_kernel, use_long_kernel, percent_short_dispersal)
NaturalAnthropogenicDispersalKernel(natural_kernel, anthropogenic_kernel, use_anthropogenic_kernel, percent_natural_dispersal)
PrimarySecondaryDispersalKernel(primary_kernel, secondary_kernel, use_secondary_kernel, percent_primary_dispersal)
DoubleDispersalKernel(first_kernel, second_kernel, use_second_kernel, percent_first_dispersal)
```

A separate, yet related, issue is parametrize usage of the kernels. Now it is either first one or both, alternative is supporting not using the first one. We could also have the types (the enum) as a parameter instead of boolean (now called use_long_kernel).

This is mostly about API beauty and it should be easy to keep backwards compatibility, so there is no pressure.

PR #61, opened 2020-05-26
addressing #50 and #30

malawrim:
changed total_plants to total_hosts in simulation
renamed short_long_kernel to natural_anthropogenic_kernel and related parameter names


malawrim:
For #30 I only renamed the parameters, I was not sure how to address the second part of the issue (parametrize usage of the kernels) because I did not fully understand the issue. Are you wanting multiple ways to call the function? (i.e one excludes natural and the other excludes anthropogenic?)

malawrim:
For #50 the only major issue I saw was with total_plants, however I also considered changing pesticideTreatment class name to be more general (i.e chemicalTreatment) because I am not sure pesticides can be related to animals. But, I was not sure about this, so I can make/not make this change based on your input.

wenzeslaus:
Great, this looks good, but I have to look in more detail for the review.

@malawrim Can you please remove the `.DS_Store` file? Simple `git rm` (+commit & push) will be enough (we will Squash it here, so the history will be clean). You can add `.DS_Store` to `.gitignore` and add that to this PR if you want.

wenzeslaus:
> I was not sure how to address the second part of the issue (parametrize usage of the kernels)

That was something to keep in mind if there is a need for it, but so far we don't have a need to generalize that, e.g., use only the second kernel. This can be also addressed on different levels including user code, so we don't have to worry about this now in the library.

wenzeslaus:
> ...I also considered changing pesticideTreatment class name to be more general (i.e chemicalTreatment) because I am not sure pesticides can be related to animals.

Good point, but I don't have any strong opinion either, let's change it once the other use cases become clear.

Thanks for catching the typos @malawrim.

PR #62, opened 2020-05-27
Optionally remove stochasticity in host movement

malawrim:
Began working on this issue by adding a non-stochastic option to simulation.movement in order to allow using a constant for movement instead of the random number generator.


malawrim:
The way I changed the movement module was to use a new parameter (movement_stochasticity) which allows the user to input 0 for the stochastic option or any other positive integer to set the infected_moved value. This may be too simplistic, so I can try something different if this seems incorrect.

ChrisJones687:
Thanks, @malawrim I hadn't thought of making the movement module non-stochastic but I like the idea. I think the simple solution here is if stochasticity in the movement module is off it should just move infected mean. #62 was meant to add a non-stochastic dispersal kernel to the disperse function. This will be a bit complicated as currently, it treats each individual that is produced in a cell individually and draws its direction and distance. Ultimately we would want a non-stochastic version to treat all dispers in a cell at once (or at least have the kernel know how many it placed where (i.e. it should calculate which cell is the most likely to have spread occur and assign the dispersed individuals to the highest probability cells first.

simulation.hpp
```diff
@@ -115,6 +115,7 @@ class Simulation
     RasterIndex cols_;
     bool dispersers_stochasticity_;
     bool establishment_stochasticity_;
+    unsigned movement_stochasticity_;
```
ChrisJones687, line 118: This could be a boolean

simulation.hpp
```diff
@@ -372,7 +379,6 @@ class Simulation
             for (int j = 0; j < cols_; j++) {
                 if (dispersers(i, j) > 0) {
                     for (int k = 0; k < dispersers(i, j); k++) {
-
                         std::tie(row, col) = dispersal_kernel(generator_, i, j);
```
ChrisJones687, line 382: This is where I was thinking of making the change but I think it would be a new type of dispersal kernel that would be called (there might be a simpler way to do this). The new dispersal kernel would need to assign either all dispers at once or remember where it assigned the previous dispers.

simulation.hpp
```diff
@@ -237,17 +241,20 @@ class Simulation
...
             }
             if (infected(row_from, col_from) > 0 && susceptible(row_from, col_from) > 0) {
-                inf_ratio = double(infected(row_from, col_from)) / double(total_plants(row_from, col_from));
+                inf_ratio = double(infected(row_from, col_from)) / double(total_hosts(row_from, col_from));
                 int infected_mean = total_hosts_moved * inf_ratio;
                 if (infected_mean > 0) {
-                    std::poisson_distribution<int> distribution(infected_mean);
-                    infected_moved = distribution(generator_);
+                	std::poisson_distribution<int> distribution(infected_mean);
+                	infected_moved = movement_stochasticity_;
```
ChrisJones687, line 254: Again good catch that this is another area to add a non-stochastic version. Here the infected mean could be moved as the non-stochastic option instead of drawing from a Poisson distribution.

malawrim: Thank you for the corrections. I will make those changes now.

malawrim:
Thank you for the help, @wenzeslaus. I knew there was a problem with the commits but was not sure how to fix it. I believe it is correct now.

simulation.hpp
wenzeslaus: Given what is in this PR now and our unclear idea about the new dispersal kernel, the kernel should be a different PR from this one. The kernel can be and practically is still the same object in our code, so indeed it can keep some state information through the different calls. We may need to video call to brainstorm this.

ChrisJones687: I agree that a video call would be good for this.

simulation.hpp
```diff
@@ -136,20 +137,23 @@ class Simulation
...
      */
     Simulation(unsigned random_seed,
                RasterIndex rows,
                RasterIndex cols,
                ModelType model_type = ModelType::SusceptibleInfected,
                unsigned latency_period = 0,
                bool dispersers_stochasticity = true,
-               bool establishment_stochasticity = true
+               bool establishment_stochasticity = true,
+               unsigned movement_stochasticity = 0
```
ChrisJones687, line 149: should also be a bool with default true

simulation.hpp
```diff
@@ -136,20 +137,23 @@ class Simulation
      * @param cols Number of columns
      * @param dispersers_stochasticity Enable stochasticity in generating of dispersers
      * @param establishment_stochasticity Enable stochasticity in establishment step
+     * @param movement_stochasticity Enable stochasticity in movement if equal to zero, other values determine infected movement
```
wenzeslaus, line 140: This needs update to bool.

malawrim: Just updated

wenzeslaus:
I was waiting for #66 to complete before merging this, but at this point it seems that #66 is almost complete. Although we don't have an immediate use, it might be good to have this functionality to have the non-deterministic options everywhere, right? Should I go ahead and merge this @ChrisJones687 ?

ChrisJones687:
Yes, merging this now and allow everything to be completely deterministic makes sense to me. Feel free to merge when convenient.

PR #66, opened 2020-06-05
Created new deterministic dispersal kernel

malawrim:
I have implemented this such that when the deterministic kernel is initialized the probability matrix is created for the entire raster. Then whenever the function is called from Simulation.disperse it creates a smaller matrix (moving window) from the probability matrix. From this, it assigns dispersers as it is called by Simulation and alters the probabilities of each cell in the window matrix so that it remembers where all of the other dispersers are. I also included functions to calculate probabilities based on the Exponential and Cauchy distributions.

Potential problems:

-  I made the moving window have a size related to the number of dispersers that the cell has, but, I can change this to anything else.

- I determined distance by calculating the difference between the row/col values of any cell to the center cell, but I am not sure if this was how you wanted to calculate the distance.

- The way the probability matrix is set up always throws dispersers towards center of the matrix even if the cell of interest is on the far edge of the matrix. For example, if the window is at the far edge of the raster (i.e cell (1,1)) right now it would say that it is more likely for the disperser to go in cell (3,3) because it is closer to the center, when it seems more likely that it would be closer to (1,1). I was not sure if this was what you were expecting or if I misunderstood how to create the probability matrix.

### DIFF
--- a/deterministic_kernel.hpp
+++ b/deterministic_kernel.hpp
@@ -16,11 +16,9 @@
 #ifndef POPS_DETERMINISTIC_KERNEL_HPP
 #define POPS_DETERMINISTIC_KERNEL_HPP
 
-#include "kernel_types.hpp"
-#include "radial_kernel.hpp"
-
 #include <vector>
 #include <cmath>
+#include <tuple>
 
 #ifndef M_PI
 #define M_PI 3.14159265358979323846
@@ -31,22 +29,30 @@
 
 namespace pops {
 
+using std::pow;
+using std::tan;
+using std::atan;
+using std::exp;
+using std::log;
+using std::ceil;
+using std::abs;
+
 /*! 
  * Cauchy distribution
  * Includes probability density function and cumulative distribution function
  * pdf returns the probability that the variate has the value x
  * cdf returns the upper range that encompasses x percent of the distribution (e.g for 99% input .99)
  */
-class cauchy_distribution
+class CauchyDistribution
 {
 public:
-	cauchy_distribution(double scale, double locator)
+	CauchyDistribution(double scale, double locator)
         : s(scale), t(locator)
     {}
 
     double pdf(double x)
     {
-    	return 1 / ((s*PI)*(1 + (std::pow((x - t)/s, 2))));
+    	return 1 / ((s*PI)*(1 + (pow((x - t)/s, 2))));
     
     }
     
@@ -54,9 +60,9 @@ public:
     {
     	if ( s == 1 && t == 0) {
     		// checking upper half so multiply result by 2
-    		return std::tan((x / 2) * M_PI) * 2;
+    		return tan((x / 2) * M_PI) * 2;
     	} else {
-    		return (std::tan((x/2)*M_PI + (std::atan(-t/s))) * s + t) * 2;
+    		return (tan((x/2)*M_PI + (atan(-t/s))) * s + t) * 2;
     	}
     }
 private:
@@ -72,24 +78,24 @@ private:
  * pdf returns the probability that the variate has the value x
  * cdf returns the upper range that encompasses x percent of the distribution (e.g for 99% input 0.99)
  */
-class exponential_distribution
+class ExponentialDistribution
 {
 public:
-exponential_distribution(double scale)
+ExponentialDistribution(double scale)
         : beta(scale)
     {}
     // assumes mu is 0 which is traditionally accepted
     double pdf(double x)
     {
-    	return (1/beta)*(std::exp(-x/beta));
+    	return (1/beta)*(exp(-x/beta));
     }
     
     double cdf(double x)
     {
     	if ( beta == 1) {
-    		return -std::log(1-x);
+    		return -log(1-x);
     	} else {
-    		return -beta * std::log(1-x);
+    		return -beta * log(1-x);
     	}
     }
 private:
@@ -131,14 +137,14 @@ protected:
     double max_distance;
     std::vector<std::vector<double>> probability;
     std::vector<std::vector<double>> probability_copy;
-    cauchy_distribution cauchy;
-    exponential_distribution exponential;
+    CauchyDistribution cauchy;
+    ExponentialDistribution exponential;
     double number_of_dispersers;
 public:
     DeterministicDispersalKernel(DispersalKernelType dispersal_kernel, 
     							Raster<int> dispersers, double dispersal_percentage,
-    							double ew_res, double ns_res, double distance_scale = 1,
-    							double locator = 0)
+    							double ew_res, double ns_res, double distance_scale = 1.0,
+    							double locator = 0.0)
         :
         dispersers_(dispersers),
         cauchy(distance_scale, locator),
@@ -154,23 +160,23 @@ public:
     	} else { 
     		//for expanding compatible kernel types
     	}
-    	number_of_columns = std::ceil(max_distance / east_west_resolution) * 2 + 1;
-    	number_of_rows = std::ceil(max_distance / north_south_resolution) * 2 + 1;
+    	number_of_columns = ceil(max_distance / east_west_resolution) * 2 + 1;
+    	number_of_rows = ceil(max_distance / north_south_resolution) * 2 + 1;
     	probability.resize(number_of_rows, std::vector<double>(number_of_columns));
     	probability_copy.resize(number_of_rows, std::vector<double>(number_of_columns));
     	mid_row = number_of_rows / 2;
     	mid_col = number_of_columns / 2;
-    	double sum = 0;
+    	double sum = 0.0;
     	for ( int i = 0; i < number_of_rows; i++ ) {
     		for ( int j = 0; j < number_of_columns; j++ ) {
     			double distance_to_center = 
-    			std::sqrt(std::pow((std::abs(mid_row - i) * east_west_resolution), 2) 
-    			+ pow((std::abs(mid_col - j) * north_south_resolution), 2));
+    			sqrt(pow((abs(mid_row - i) * east_west_resolution), 2) 
+    			+ pow((abs(mid_col - j) * north_south_resolution), 2));
     			// determine probability based on distance
 				if (kernel_type_ == DispersalKernelType::Cauchy) {
- 					probability[i][j] = std::abs(cauchy.pdf(distance_to_center));
+ 					probability[i][j] = abs(cauchy.pdf(distance_to_center));
  				} else if ( kernel_type_ == DispersalKernelType::Exponential ) {
- 					probability[i][j] = std::abs(exponential.pdf(distance_to_center));
+ 					probability[i][j] = abs(exponential.pdf(distance_to_center));
 				}
  				sum += probability[i][j];
     		}
@@ -191,10 +197,8 @@ public:
      *  Selects next row/col value based on the cell with the highest probability
      *  in the window.
      *
-     *  Generator is not used
      */
-    template<typename Generator>
-    std::tuple<int, int> operator() (Generator& generator, int row, int col)
+    std::tuple<int, int> spread(int row, int col)
     {
 		// reset the window if considering a new cell
     	if ( row != prev_row || col != prev_col ) {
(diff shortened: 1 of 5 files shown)


malawrim:
I have a lot of comments in the code right now that are mostly just notes for myself, so I will remove those before the changes are accepted.

ChrisJones687:
Overall this is a great start. Thanks for working on this. 
A couple of key observations:
1. The probability matrix should be the moving window and account for ~99% of all possible dispersal (we might want to set up a call to discuss this). If our full raster is 1000 x 1000 the probability matrix might only be 5x5. 

- The moving window shouldn't change for each cell. It should be set at the beginning of the simulation, based on having ~99% (we might need to make this as low as 95) of all dispersal events occur within that window based on the dispersal kernel that was selected for the simulation. 

- It should be calculated from the center of the cell of the moving window (probability matrix) to the center of each cell in the probability. It should also take into account the spatial resolution. 

- Ideally, this should still allow for dispersers outside of the main raster if called from the edges. 

Let me know if a video chat would be helpful. I haven't had a chance to review the code yet but from the explanation, I think the main thing that needs to change is the probability matrix being the moving window, not the entire landscape. It sounds like the rest of the code base that you have here should still work to return the desired results.

malawrim:
I am not sure I understand the "account for ~99% of all possible dispersal" part. Would the moving window size just be determined by selecting cells based on the distribution? For example, if the distribution shows that 99% of the dispersers would fall within 2 meters you would select cells that are in that range to be the moving window. If this is not correct, a video call would be helpful to discuss it in more detail.

ChrisJones687:
@malawrim Yes the moving window would be determined by selecting cells based on the distribution and that if 99% of all dispersal falls within 2 meters then we would select an n x n window (n is odd) that allows for this to happen based on that distance and the spatial resolution of the cells. This will mean a larger window when we use a Cauchy distribution than when we use an exponential but that makes sense based on the distribution shapes.

malawrim:
I have changed the implementation to create a probability matrix when the kernel is initialized, the size of this window is determined by calling a new function (one for Cauchy and another for Exponential). This function will return the minimum row/column value the window should have so that it will include any specified percentage (i.e 99%) of possible dispersal. I determined the function to use for this function by modifying the cumulative distribution function (expressed as an integral) and set the equation equal to a variable instead of 1 in order to find a certain percentage of the distribution. I then simplified and rewrote the equation to solve for the x(max) value, so given a certain percentage it returns value of the x-axis at this point on the distribution.

Notes:
I am concerned about the difference in the size of the windows for exponential vs cauchy. While I understand exponential should be smaller, when I account for spatial resolution, exponential can end up being 1x1, which does not work. I am not sure if I misunderstood how to appropriately account for spatial resolution or if I have had some miscalculation in calculating the distribution.

deterministic_kernel.hpp
```diff
@@ -0,0 +1,289 @@
...
+    				max = probability_copy[i][j];
+    				max_prob_row = i;
+					max_prob_col = j;
+    			}
+    		}
+    	}
+
+		// subtracting 1 preserves the probability order but ensures that this cell
+		// is not select again unless all other cells in window have been selected
+    	probability_copy[max_prob_row][max_prob_col] -= 1;
```
ChrisJones687, line 263: It should subtract 1- (1/number_of_dispersers). The 1/number of dispersers should be initialized with the probability matrix based on the value in the disper matrix on the first pass (i.e. if row and column are new get the value from the dispersers matrix otherwise 1/ number of dispersers should be used for the rest of the dispersers from that cell to modify the probability. Doing it this way ensures we always move the same proportion of the individuals to each cell no matter how many are dispersing.

test_simulation.cpp
```diff
@@ -171,6 +172,164 @@ int test_with_reduced_stochasticity()
...
+        cout << "Deterministic Kernel Exponential: infected (actual, expected):\n" << infected << "  !=\n" << expected_infected << "\n";
+        return 1;
+    }
+    if (mortality_tracker != expected_mortality_tracker) {
+        cout << "Deterministic Kernel Exponential: mortality tracker (actual, expected):\n" << mortality_tracker << "  !=\n" << expected_mortality_tracker << "\n";
+        return 1;
+    }
+    cout << "Deterministic Kernel Exponential: no errors\n";
+    
+    // testing cauchy pdf & cdf
```
wenzeslaus, line 280: This is separate from the rest, so make it a separate function.

In fact, it would be ideal if this would be in a separate file. However, there is couple of things which need to be done for that (see Makefile), so I can do it later.

test_simulation.cpp
```diff
@@ -171,6 +172,164 @@ int test_with_reduced_stochasticity()
...
+    double scale = 0.001;  // rounding to thousands place
+    CauchyDistribution cauchy(1.0, 0.0);
+    double probability = (int) (cauchy.pdf(5) / scale) * scale;
+    if ( probability != 0.012) {
+    	cout << "Cauchy Distribution: probability was " << probability << " but should be 0.012\n";
```
wenzeslaus, line 285: `0.012` will be better as a variable something like `probability_ref`(erence). Then you can also do `<< probability_ref`. More characters, but only one place where the number is. (Same for the others.)

deterministic_kernel.hpp
```diff
@@ -0,0 +1,246 @@
...
+    /*! Generates a new position for the spread.
+     *
+     *  Creates a copy of the probability matrix to mark where dispersers are assigned.
+     *  New window created any time a new cell is selected from simulation.disperse
+     *
+     *  Selects next row/col value based on the cell with the highest probability
+     *  in the window.
+     *
+     */
+    std::tuple<int, int> spread(int row, int col)
```
wenzeslaus, line 201: This could be a function call operator as it is in the other kernels (just copy-paste the syntax and then delete `.spread` from the place where you call it).

With how you designed your code, you are not using this kernel directly as the other kernels are used, so it is not necessary, but using the operator would be more consistent.

deterministic_kernel.hpp
```diff
@@ -0,0 +1,246 @@
...
+    // maximum distance from center cell to outer cells
+    double max_distance;
+    std::vector<std::vector<double>> probability;
+    std::vector<std::vector<double>> probability_copy;
+    CauchyDistribution cauchy;
+    ExponentialDistribution exponential;
+    double number_of_dispersers;
+public:
+    DeterministicDispersalKernel(DispersalKernelType dispersal_kernel, 
+    							Raster<int> dispersers, double dispersal_percentage,
```
wenzeslaus, line 145: The type of dispersers needs to be little more complicated:

1. It should be a const reference, i.e., `const ... &` so that the data does not have to be copied, but it is guaranteed in the API that we don't modify the data.
2. The kernel class actually needs to be a class template with one template parameter IntegerRaster, see, e.g., the Simulation class. This allows us to use different raster/matrices/grids in the different APIs. If this turns out to be too much hassle, I can add this generalization later.

deterministic_kernel.hpp
```diff
@@ -0,0 +1,246 @@
...
+    // position of cell from previous call
+    int prev_row = -1;
+    int prev_col = -1;
+    // number of rows/cols in the probability window
+    int number_of_rows = 0;
+    int number_of_columns = 0;
+    // maximum distance from center cell to outer cells
+    double max_distance;
+    std::vector<std::vector<double>> probability;
+    std::vector<std::vector<double>> probability_copy;
```
wenzeslaus, line 139: If it would help, these two actually could be `Raster<double>`. We are using the raster class (template) when there is no other fitting option (in R we are using NumbericMatrix), however we could reuse Raster class for cases such as this one when using the native data structures does not make sense and we need some kind of 2D numeric array. Using Raster could make some of the operations simpler.

test_simulation.cpp
malawrim: I moved this to a different function, but can move it to a different file later if you would like.

wenzeslaus: Well, that would be actually quite useful because the test_simulation.cpp is getting long. There is no doc for adding more test files, so let me know if I should clarify something.

malawrim: I created a new test file test_deterministic.cpp and I moved the deterministic kernel tests to that file. I also updated the makefile.

deterministic_kernel.hpp
```diff
@@ -0,0 +1,247 @@
...
+                }
+            sum += probability(i,j);
+            }
+        }
+        //normalize based on the sum of all probabilities in the raster
+        for ( int i = 0; i < number_of_rows; i++ ) {
+            for ( int j = 0; j < number_of_columns; j++ ) {
+                probability(i,j) /= sum;
+            }
+        }
```
wenzeslaus, line 193: `probability /= sum;` will be now sufficient with Raster and hopefully equally readable. Just trying to save lines.

malawrim:
I changed the formula for the cauchy cdf because I found an error. I also changed the names of the functions from cdf to icdf because the rewritten form of the cdf I am using is called the inverse cumulative distribution function (or quantile function).

deterministic_kernel.hpp
```diff
@@ -0,0 +1,247 @@
...
+ * dispersal_percentage is the percent of all possible dispersal to be included
+ * in the moving window size (e.g for 99% input 0.99).
+ *
+ * Useful for testing as it is deterministic and provides fully replicable results
+ */
+template<typename IntegerRaster>
+class DeterministicDispersalKernel
+{
+protected:
+    IntegerRaster dispersers_;
```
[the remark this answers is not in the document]
malawrim: Since IntegerRaster dispersers is passed into the RadialDispersalKernel constructor, I also changed that to be const, but if this is wrong I can change it back.

test_simulation.cpp
```diff
@@ -27,6 +27,7 @@
 #include "neighbor_kernel.hpp"
 #include "radial_kernel.hpp"
 #include "simulation.hpp"
+#include "deterministic_kernel.hpp"
```
wenzeslaus, line 30: No longer needed? Maybe same applies to other changes in this file?

radial_kernel.hpp
```diff
@@ -175,12 +177,18 @@ class RadialDispersalKernel
     std::cauchy_distribution<double> cauchy_distribution;
     std::exponential_distribution<double> exponential_distribution;
     von_mises_distribution von_mises;
+    DeterministicDispersalKernel<IntegerRaster> deterministic_kernel;
+    bool deterministic_;
 public:
     RadialDispersalKernel(double ew_res, double ns_res,
                           DispersalKernelType dispersal_kernel,
-                          double distance_scale,
+                          double distance_scale = 1.0,
```
ChrisJones687, line 185: Why add a default here? It should be a required parameter that we need from the user (or the calibration).

radial_kernel.hpp
```diff
@@ -175,12 +177,18 @@ class RadialDispersalKernel
...
                           Direction dispersal_direction = Direction::None,
-                          double dispersal_direction_kappa = 0
+                          double dispersal_direction_kappa = 0,
+                          bool deterministic = false,
+                          const IntegerRaster& dispersers = {{0}},
+                          double dispersal_percentage = 0.99,
+                          double locator = 0.0
```
ChrisJones687, line 191: We may want to just have this hard-coded to zero in the distribution (thoughts @wenzeslaus) and not a parameter as we always want the location to be the center of the cell, which would make it more consistent with the current implementation of the Cauchy distribution.

deterministic_kernel.hpp
```diff
@@ -0,0 +1,240 @@
...
+     *  Selects next row/col value based on the cell with the highest probability
+     *  in the window.
+     *
+     */
+    template<class Generator>
+    std::tuple<int, int> operator()(Generator& generator, int row, int col)
+    {
+        // reset the window if considering a new cell
+        if ( row != prev_row || col != prev_col ) {
+            number_of_dispersers = 1.0 / (double)dispersers_(row, col);
```
ChrisJones687, line 203: rename number_of_dispersers to proportion_of_dispersers to be more precise in the naming and easier to follow

radial_kernel.hpp
malawrim: I had the distance_scale and locator values set up so it would default to the standard cauchy/exponential distribution, but I can take it out.

radial_kernel.hpp
wenzeslaus: I'm not sure right now what is the right place, but +1 for hardcoding, so it is always a center of the cell.

malawrim: Would you also like for me to remove the locator parameter from the deterministic kernel constructor and the CauchyDistribution pdf and icdf functions?

ChrisJones687:
Yes, so that it is consistently 0.0 throughout since we don't allow the
user to change this value. Thanks!

ChrisJones687:
Merge when you are ready. I forgot if this was next inline or if it was the Model PR. Thanks!

PR #76, opened 2020-06-29
Use consistent formatting with clang-format

wenzeslaus:
This reformats the code using clang-format and will add a configuration to help ensure keeping of this format (WIP).


wenzeslaus:
@malawrim @petrasovaa Please just see if there are any major oversights in the style by checking the code here and there. I had to do quite a bit of customization.

The biggest known issue is formatting conditions in if statements and how they align with other code. The current choices in clang-format don't allow to resolve it without completely moving from the idea to make it somehow similar to the broadly accepted formatting in Python we are using. So, the question is if you think it is extremely bad.

Issue #65, opened 2020-06-04
Add quarantine escape functionality


petrasovaa:
Add functionality to compute the following metrics:

* Probability of escaping quarantine
* Year of escape
* Distance to quarantine

Possible implementation (in statistics.hpp):
* New class for quarantine
  * raster input representing quarantine area(s) (0 - no quarantine, 1, 2, 3 ... separate quarantine areas)
  * finds N, S, E, W boundary of the area(s)
  * method that gets infected cell raster and for each infected cells finds out out if inside area, and if yes, computes distance in 4 directions to the boundary. Reports if inside or not and minimum distances in all directions (or just the direction with the minimum distance).
  * aggregate across all quarantine areas (alternatively report separately)
* Function to aggregate the results of this method for multiple runs to compute the probability of escape, report the distribution of min distances.

PR #84, opened 2020-07-17
Quarantine escape metrics

petrasovaa:
This PR addressing issue #65 adds the functionality to compute probability of escaping quarantine for a certain year and the (shortest) distance to quarantine border (distance to quarantine border bbox).

The implementation is very similar to spread rate computation.

The input can contain multiple quarantine areas as raster (0 means no area, any integer serves as a category of the individual areas). The shortest distance is aggregated over all the areas. This could be changed to track shortest distances for each area but will add complexity.

The usage of the class is integrated into the new model interface. Currently, the implementation assumes escaping quarantine is checked at the end of each year. The implementation includes a function to write out results for all years and runs as a string, includes information about probability of infection for that year and list of min distances for each run.


petrasovaa:
> Overall great job. I will need to look into how to handle the tuple output in Rcpp as it doesn't (or didn't) handle exporting tuples from C++ to R. I have a wrapper that does that for spread rates by converting into an array but that isn't going to work in this case since they aren't all ints.

Let me know if you need to change the interface to work better with Rcpp then.

ChrisJones687:
> > Overall great job. I will need to look into how to handle the tuple output in Rcpp as it doesn't (or didn't) handle exporting tuples from C++ to R. I have a wrapper that does that for spread rates by converting into an array but that isn't going to work in this case since they aren't all ints.
> 
> Let me know if you need to change the interface to work better with Rcpp then.

The only aspect that needs to change is the return, could we use list for that instead?

petrasovaa:
I added methods `escaped`, `distance` and `direction` for simpler access in rpops. I removed 'year' and replaced with 'step' to make it more generic, however I didn't change the schedule creation in config.hpp because I was not sure how you want to do it (have separate quarantine_frequency and quarantine_frequency_n parameters like we have for output?)

ChrisJones687:
> 
> 
> I added methods `escaped`, `distance` and `direction` for simpler access in rpops. I removed 'year' and replaced with 'step' to make it more generic, however I didn't change the schedule creation in config.hpp because I was not sure how you want to do it (have separate quarantine_frequency and quarantine_frequency_n parameters like we have for output?)

Yes I think that makes the most sense ( to use frequency and frequency_n) and is probably how we should handle all schedules for maximum flexibility. This will require a few minor changes to some of the other functions namely mortality but doesn't affect this one!

petrasovaa:
I added the frequency there. Function `output_schedule_from_string` should be renamed to `schedule_from_string`, I was not sure if it doesn't interfere with your modifications for mortality, so I didn't touch it.

ChrisJones687:
> I added the frequency there. Function `output_schedule_from_string` should be renamed to `schedule_from_string`, I was not sure if it doesn't interfere with your modifications for mortality, so I didn't touch it.

This can be changed as well it won't affect my changes to the mortality flexibility.

PR #86, opened 2020-07-29
Expanding functionality of model api

malawrim:
Added deterministic_kernel and simulation.movement to model/config

### DIFF
--- a/include/pops/config.hpp
+++ b/include/pops/config.hpp
@@ -43,6 +43,8 @@ public:
     // Reduced stochasticity
     bool generate_stochasticity{true};
     bool establishment_stochasticity{true};
+    bool movement_stochasticity{true};
+    bool deterministic{false};
     double establishment_probability{0};
     // Temperature
     bool use_lethal_temperature{false};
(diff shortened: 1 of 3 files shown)


include/pops/config.hpp
```diff
@@ -70,6 +72,11 @@ class Config
     bool use_mortality{false};
     double mortality_rate{0};
     int first_mortality_year{0};  // TODO: document that it starts at 1, not 0
+    unsigned last_index{0};
+    bool use_movements{false};
+    std::vector<std::vector<int>> movements;
```
wenzeslaus, line 77: I think `movements` should be a parameter of run_step() function like the data are and the things like Treatments. `last_index` should be part of the Module class because it is something what is changing during simulation.

The current design of movement is not perfect, I think for the future we may try to come up with something like Treatments class for the movements so that things like last_index are encapsulated there and run_step itself is more flexible.

ChrisJones687: Yes, I agree with Vashek on the movement class needing to be changed long-term. I have added #87 for that as a future improvement.

include/pops/model.hpp
```diff
@@ -114,9 +82,44 @@ class Model
         Treatments<IntegerRaster, FloatRaster>& treatments,
         IntegerRaster& resistant,
         std::vector<std::tuple<int, int>>& outside_dispersers,  // out
-        SpreadRate<IntegerRaster>& spread_rate  // out
-    )
+        SpreadRate<IntegerRaster>& spread_rate,  // out
+        std::vector<std::vector<int>> movements)
```
wenzeslaus, line 86: Just a detail: It would be better if the parameter is a const reference (`const std::vector<std::vector<int>>&`) like in `Simulation::movement()`. We don't need a copy here and we don't modify it as output either. Even if there would be no performance difference with the copy, cost ref better documents the intention here.

wenzeslaus:
You can just merge afterwards!

PR #89, opened 2020-08-12
Bugfix/88

### LINKED ISSUE #88: Exposed classed not being managed with Treatment
Currently, the exposed class is not being removed during treatment. This should be an easy fix using treatment.manage_mortality and I have a version of this running locally in rpops and will push here once I confirm it works. We may want to rename treatments.manage_mortality but don't have a good solution.

ChrisJones687:
Added loop to remove exposed class. Fixes #88. This is a very easy solution with the current implementation and is incredibly flexible this way. We probably want to rename treatments::manage_mortality but I don't have a good name for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project should be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## 2020-08-12
+- Exposed class now being removed with treatments
+  * Added exposed class treatments to model class
+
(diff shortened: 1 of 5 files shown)


wenzeslaus:
This should be done, but want to add tests before merging to be sure.

wenzeslaus:
This has a test now and I also fixed the bug I had there (!).

ChrisJones687:
> This has a test now and I also fixed the bug I had there (!).

This looks good to me. Thanks for moving it into the treatment functionality. This is a much better implementation.

PR #97, opened 2020-08-24
Instead of a vector pass just reference to weather_coefficient

wenzeslaus:
Since currently the weather coefficient is handled by the caller rather than as part of the model,
the decision which one to use is based done outside of Model::run_step().
This way rpops can keep full list of wather coefficients while r.pops.spread can
dynamically load only what is needed (for the current output step).


include/pops/model.hpp
```diff
@@ -141,7 +141,7 @@ class Model
                 dispersers,
                 infected,
                 config_.weather,
-                weather_coefficients[weather_step],
```
ChrisJones687, line 144: if possible weather_step could be removed from line 73.

wenzeslaus: Huh, I relied on compiler warnings, but forgot I did not enable them for CMake. Thanks for catching that.

Issue #91, opened 2020-08-18
total_hosts in Model is actually not total hosts, it is "all different individuals"


wenzeslaus:
I realized there was a reason I never renamed `total_plants` to `total_hosts`. Even in simple SI model without pest resistance, `S + I != total_plants` because the model includes a dilution effect where number of all (not only host) plants in the cell influences the establishment:

```
double probability_of_establishment =
                                (double)(susceptible(row, col)) / total_hosts(row, col);
```

https://github.com/ncsu-landscape-dynamics/PoPS/blob/master/include/pops/simulation.hpp#L419

This is in the `disperse` function.

On the other hand, the newer `movement` function assumes `S + I == total_hosts`. I don't think the limitation/assumption in `movement` necessitates that the whole model needs to require the same condition as long as it is documented that the `movement` can only be used when the condition is met.

ChrisJones687:
Would a better parameter name be total_populations? This would get past the animal and plant issue while being more correct than total hosts and then we just define in the documentation that total_populations is the cumulative hosts and non-hosts in a cell and must be >= S + E + I.

Agreed that the rest of the model doesn't need to and shouldn't mimic that assumption in the movements module. We should also make this assumption more flexible when we address #87.

wenzeslaus:
Fixed by #98.

Issue #87, opened 2020-07-31
Create movements class


ChrisJones687:
The current implementation of movements is not ideal. We would like to create a class that is more similar to treatments to allow for more flexibility.

ChrisJones687:
- [ ] Add exposed and resistant to movements #93 

- [ ] Add new spatial index location if movements occur in a location with no host from #116

PR #98, opened 2020-08-24
Rename total_hosts to total_populations and document run_step

wenzeslaus:
This renames total_hosts to total_populations (#91), updates documentation to emphasize the distinction,
and adds documentation to Model::run_step().

### DIFF
--- a/include/pops/model.hpp
+++ b/include/pops/model.hpp
@@ -81,7 +81,7 @@ public:
      * No treatment can be applied when movement is active because host movement does
      * not support resistant hosts.
      *
-     * *dispersers* is for intrnal use and for stracking dispersers creation.
+     * *dispersers* is for internal use and for tracking dispersers creation.
      * The input values are ignored and the output is not the current existing
      * dispersers, but only the number of dispersers generated (and subsequently
      * used) in this step. There are no dispersers in between simulation steps.
--- a/include/pops/model.hpp
+++ b/include/pops/model.hpp
@@ -68,12 +68,57 @@ public:
               config.movement_stochasticity)
     {}
 
+    /**
+     * @brief Run one step of the simulation.
+     *
+     * The *total_populations* can be total number of hosts in the basic case
+     * or it can be the total size of population of all relevant species
+     * both host and non-host if dilution effect should be applied.
+     * When movement is applied, *total_populations* needs to be only the
+     * total number of hosts because movement does not support non-host
+     * individuals.
+     *
+     * No treatment can be applied when movement is active because host movement does
+     * not support resistant hosts.
+     *
+     * *dispersers* is for intrnal use and for stracking dispersers creation.
+     * The input values are ignored and the output is not the current existing
+     * dispersers, but only the number of dispersers generated (and subsequently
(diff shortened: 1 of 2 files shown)


wenzeslaus:
@ChrisJones687 Let me know if total_populations is what you meant. The plural there seems to me a little strange.

include/pops/model.hpp
```diff
@@ -68,12 +68,57 @@ class Model
...
+     * or it can be the total size of population of all relevant species
+     * both host and non-host if dilution effect should be applied.
+     * When movement is applied, *total_populations* needs to be only the
+     * total number of hosts because movement does not support non-host
+     * individuals.
+     *
+     * No treatment can be applied when movement is active because host movement does
+     * not support resistant hosts.
+     *
+     * *dispersers* is for intrnal use and for stracking dispersers creation.
```
ChrisJones687, line 84: typos  internal and tracking

include/pops/simulation.hpp
```diff
@@ -215,11 +215,15 @@ class Simulation
...
+     *
+     * @note Note that unlike the other functions, here, *total_hosts*,
+     * i.e., number of hosts is required, not number of all hosts
+     * and non-host individuals.
      *
      * @param infected Currently infected hosts
      * @param susceptible Currently susceptible hosts
      * @param mortality_tracker Hosts that are infected at a specific time step
-     * @param total_hosts total number of hosts
+     * @param total_hosts Total number of hosts
```
ChrisJones687, line 226: total_populations

ChrisJones687: We either need to change it everywhere and just make it explicitly clear (which you have done) that total_populations can only include hosts when movement is used. Otherwise, we need to use 2 parameters (total_hosts and total_populations) just to support this aspect of movements in the near-term until the movements class is redone in #87. I am open to either option just think changing it is easier for handling the data in R and Grass but it does put the onious on the user to pay attention to the documentation here.

ChrisJones687: Actually, this would be a good first segway into handling the muli-host dynamics with host competency and susceptibility.

include/pops/model.hpp
wenzeslaus: Fixed, thanks.

include/pops/simulation.hpp
wenzeslaus: I actually meant for this to be total_hosts. My thinking here was to keep the naming as whatever it actually is, so total_hosts. The parameter names are not part of API in C++, so they are more documentation and for clarity of the code. Also, this is internal to core, so it should not be visible to rpops and r.pops.spread. Hence, total_hosts here, total_populations elsewhere. I think it should be changed everywhere in rpops and r.pops.spread since what I'm saying does not apply there. On the other hand, the point was to document it well which includes avoiding confusion which is perhaps what I created.

ChrisJones687: That makes complete sense now! It makes it clear when we go to update the movements class that this is the assumption here. Good call. Merge when ready!

PR #100, opened 2020-08-26
Enable compiler warnings in CMake

wenzeslaus:
This fixes number of C++ warnings from GCC (and clang), namely:

* Unused variables and parameters (some removed, some fixed, some marked as unused).
* pops::Raster now uses int by default as index instead of unsigned int to match Rcpp. (Applies in test.)
* Default assignment operator needs to be defined explicitly.
* Catching exceptions in a wrong way.


malawrim:
Looks good, there were no warnings when I ran it on my computer.

Issue #101, opened 2020-08-26
Update Spread rate timing


ChrisJones687:
Add the capability for spread rates to be calculated at every output interval rather than just yearly. Should be able to be handled in config similarly to quarantine stats. We should default to handling all of our statistical outputs this way and add a use_spreadrates (or whatever stat) bool to turn off the functionality if we don't need to use it for that case study.

PR #103, opened 2020-08-28
Feature/flexible spread rates

ChrisJones687:
Update to spread rate schedule to make the timing more flexible similar to quarantine.


ChrisJones687:
closes #101

PR #109, opened 2020-09-04
Add a separate section on how to cite PoPS and clarify authors

wenzeslaus:
This is a new How to cite section (the two papers to be replaced with Jones 2020) and
restructured Authors and contributors section. Both are to be fully synced with
rpops and r.pops.spread.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,48 @@ and pathogens in forest and agricultural landscapes.
 It is used for various pest, pathogens, and hosts.
 It was originally developed for *Phytophthora ramorum* and the original
 version of the model was written in R, later with Rcpp,
-and was based on this reference paper:
-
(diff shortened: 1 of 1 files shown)


wenzeslaus:
We could make it slightly different for each repo, e.g., to include a more specific example of acknowledging or citing the software (no the example is for rpops as the most likely one to be used).

I went ahead and created PR for the other two repos, so you can review the changes in light of how they will look in the other repos.

https://github.com/ncsu-landscape-dynamics/rpops/pull/37
https://github.com/ncsu-landscape-dynamics/r.pops.spread/pull/33

ChrisJones687 (approved):

ChrisJones687:
looks good.

Issue #28, opened 2019-08-13
Implement additional kernels


wenzeslaus:
Based on comment by @ChrisJones687 in #19:

> ### Kernels to implement
> ([comes from data on seed dispersal but very applicable](https://besjournals.onlinelibrary.wiley.com/doi/10.1111/1365-2745.12666))
> ...

#### Available in C++11

- [x] Exponential: single parameter should be 1/(mean distance (this is calculated in MCMC))
- [x] Gamma: double parameter (shape (1.0) and scale(estimated))
- [x] Cauchy: double parameter (shift (0.0) and scale(estimated))
- [x] Weibull: double parameter (shape (1.0) and scale (estimated))
- [x] Normal: double Parameter (mean (0.0) and sd (estimated))
- [x] Lognormal: double parameter (mean (0.0) and sd (estimated))

#### Available in Boost, but not available in standard C++
> Available in [boost C++ math library](https://www.boost.org/doc/libs/1_68_0/boost/math/distributions.hpp)
- [x] WALD or Inverse Gaussian
- [x] Logistic distribution
> 
#### Others to be implemented, not in C++11 or Boost
- [x] 2Dt
- [x] Power Law (Cauchy is one variant of a power law)
- [x] log hyperbolic secant
- [x] Exponential Power

#19 is now resolved by #27, so this can be implemented.

PR #110, opened 2020-09-07
Adding additional kernels from issue #28

malawrim:
Added the following kernels:
- Power Law
- Hyperbolic secant
- Exponential Power
- Weibull
- Normal
- Lognormal
- Gamma
- Logistic


wenzeslaus:
The Makefile was deleted in master branch, so deleting it here should resolve the conflict for that file.

The new files need a new line at the end of file (EOL at EOF). Apparently, clang-format does not check for that and considers that something to do/check by text editors or other tools (for example Git warns about that in diff, GitHub shows there some kind of warning sign).

malawrim:
I removed the location parameter from the kernels, but some of the distributions still use two parameters (scale and shape). Would you like to include the shape parameter or remove this as well?

wenzeslaus:
> I removed the location parameter from the kernels, but some of the distributions still use two parameters (scale and shape). Would you like to include the shape parameter or remove this as well?

I think we should keep shape. As far as I understand, you can really make a proper use of distributions like gamma unless you a) use shape or b) you actually pick a specific shape of that distribution (and hardcode that). I don't think we want to go with option b) because it is less flexible and not really taking full advantage of these distributions.

ChrisJones687:
> > I removed the location parameter from the kernels, but some of the distributions still use two parameters (scale and shape). Would you like to include the shape parameter or remove this as well?
> 
> I think we should keep shape. As far as I understand, you can really make a proper use of distributions like gamma unless you a) use shape or b) you actually pick a specific shape of that distribution (and hardcode that). I don't think we want to go with option b) because it is less flexible and not really taking full advantage of these distributions.

Agreed, I think we should keep it as well.

include/pops/kernel_types.hpp
```diff
@@ -74,6 +82,22 @@ inline DispersalKernelType kernel_type_from_string(const std::string& text)
         return DispersalKernelType::Uniform;
     else if (text == "deterministic-neighbor" || text == "deterministic_neighbor")
         return DispersalKernelType::DeterministicNeighbor;
+    else if (text == "power law" || text == "power-law")
```
ChrisJones687, line 85: This is great. My only small suggestion and this goes for the previously implemented kernels as well (something I should have thought of earlier) is to make it so that we also account for capitalization (e.g. include both "gamma" and "Gamma").

ChrisJones687 (approved):

include/pops/power_law_kernel.hpp
```diff
@@ -0,0 +1,98 @@
...
+    template<class Generator>
+    double random(Generator& generator)
+    {
+        std::uniform_real_distribution<double> distribution(0.0, 1.0);
+        double x = distribution(generator);
+        // return pow(x, (1.0 / (-alpha_ + 1.0))) * xmin_;
+        // since distribution is 0 to 1 xmin needs to be greater than 0
+        if (xmin <= 0) {
+            xmin = 0.01;
+        }
```
wenzeslaus, line 55: This is something to resolve once in the constructor for all functions and function calls.

Is there any compelling reason to choose 0.01 and not 0.001 or anything else? You could also throw std::invalid_argument. We usually do nice user friendly checks in the interfaces (rpops, r.pops.spread) as we are closer to what user may want or expect and here we just go with exceptions to avoid surprising results.

wenzeslaus (changes requested):
I agree with the code layout and +1 this overall. I did not check the equations of the distribution/pdf/icdf functions, but let me know if you want me to go through those as well.

This can be merged once the std::array issue is addressed. The other issues could go to the technical debt file or new issues if you don't want to deal with that now.

For the record (e.g., as a note to future myself), I agree with how test_random.cpp is implemented. In next iteration, we may want to run both the local copy and the actual implementation.

include/pops/cauchy_kernel.hpp
```diff
@@ -0,0 +1,85 @@
...
+     *  Returns random value from cauchy distribution
+     *  Used by RadialKernel to determine location of spread
+     *  @param generator uniform random number generator
+     *  @return value from cauchy distribution
+     */
+    template<class Generator>
+    double random(Generator& generator)
+    {
+        return std::abs(cauchy_distribution(generator));
+    }
```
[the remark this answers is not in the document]
malawrim: I wanted to explicitly name it random so that it would be more clear that it is used only by radial kernel while the other two functions are used by deterministic kernel. But, I can also name it something else or change it to operator() if that would be more clear.

include/pops/power_law_kernel.hpp
```diff
@@ -64,8 +69,14 @@ class PowerLawKernel
...
     double pdf(double x)
     {
-        if (x <= 0 || xmin == 0 || alpha <= 1.0) {
-            return 0;
+        if (x < 0) {
+            throw std::invalid_argument("x cannot be less than 0.0");
+        }
+        // x = distance to center and is always 0 for the center cell
+        // so have to add xmin to account for minimum x values
+        x = x + xmin;
```
malawrim, line 77: I am not sure about this addition. I think xmin functions similarly to the locator parameter so ideally we could set this to zero like all the of other locator parameters, but the function is undefined when xmin is zero. Also x needs to be greater than or equal to xmin, but x will always be zero for the center cell so this causes an error because x is less than xmin. Do you think it is better to leave in this line (i.e shift the starting distance) or set xmin to some value close to zero (i.e 0.001)?

ChrisJones687 (approved):

include/pops/power_law_kernel.hpp
malawrim: I think that this is the best solution. I realized that even if xmin is hard-coded, x still must be greater than xmin or the result is invalid. Also, I tested different values of xmin (i.e 0.001 vs 0.0001) and it strongly effects the distribution between x-values 0-2 by ~0.10.

ChrisJones687: I agree that this is the best way forward.

wenzeslaus (approved):

Issue #111, opened 2020-09-07
fails build on macOS


ChrisJones687:
Discovered during R cmd check on macOS. Need to add override to get_start and get_end functions in treatment

Issue #39, opened 2019-11-05
errors when displaying workshop scenarios done with TL


dagaydos21:
So that thing that happened during the meeting with David has been continuing. If you try to display the results from a scenario that we did at the workshop via Tangible Landscape, the "waiting for Tangible Landscape" continues to ping after you hit "stop listening". It doesn't do this with the scenarios created via the PoPS dashboard. Dashboards 97 and 98 both have the error. 

In one instance, I got an error code that says: An error has occurred: 0. Error: undefined. There was a problem listening for the results of the model run... the simulation is most likely still running. Try refreshing the page in 1-2 minutes. 

It seems like this started after the changes to the API, and after I pulled all of the management data to recalculate costs, so the error point could be one of those things.

PR #112, opened 2020-09-07
Fix macOS buid fail

ChrisJones687:


ChrisJones687:
@wenzeslaus adding the override here passes the R build process for all 3 OS put fails here on CMake. Thoughts.

wenzeslaus (approved):
As far as I understand, this was only a compilation warning in https://github.com/ncsu-landscape-dynamics/rpops/pull/39 not an error, but that's actually just depends on the choice of compiler flags. (Looks like compiler warning is a failure for the check.)

Anyway, we _do_ want the override to be there.

We need to add clang to CI here (#113) so that we get all the checks which both GCC and Clang enable with -Wall -Wextra which are apparently different in each case.

ChrisJones687:
Yes, it is a compilation warning. And a compilation warning of any kind translates to a build error in R and inability to install the package. Warnings as failures were the main reason for adding the GitHub actions so we could test builds on all 3 OS in ncsu-landscape-dynamics/rpops#39.

ChrisJones687:
Fixes #111

wenzeslaus:
Thanks for the clarification.

PR #120, opened 2020-10-05
Spatial index

ChrisJones687:
I ended up using std::vector as std::pair is a specific case of std::tuple and doesn't work as well with R.


ChrisJones687:
Finished profiling. A couple of notes. I ran the profiling on my laptop as I thought it would be more representative of server cpu speed than my desktop. Results are plotted below for 10 runs at each host level. Interestingly, on a percentage basis it gets better at higher levels of host availability and smooths out variability in run times. Not 100% sure why other than it completely eliminates the calculations in dispersers where there are no hosts rather than checking if it is there at all.


profile_comparison

malawrim (approved):
Looks good to me

wenzeslaus (changes requested):
For the naming try:

```cpp
        for (auto indices : suitable_cells) {
            int i = indices[0];
            int j = indices[1];
            auto value = quarantine_areas(i, j);
```

wenzeslaus (approved):

ChrisJones687:
closes # 116

Issue #74, opened 2020-06-23
Move the utility functions and defs to a separate file


wenzeslaus:
There is couple of general utility functions or other definitions in the code. Some wrapping or extending standard library facilities. Move these into as separate file.

Some defs many need to be in a separate file such as an index type which might be needed to at least partially resolve `-Wsign-compare` from #72.

Issue #73, opened 2020-06-23
Move the general math or stat functions to separate file or files


wenzeslaus:
Some parts of code need a general mathematical or statistical function which is not in the standard library. Move these function out of the current files into one or more files with general functionality.

Issue #29, opened 2019-08-14
Date class: make conversion to string easy


petrasovaa:
This is needed for C/C++ for printing debug messages (in r.pops.spread).

PR #121, opened 2020-10-09
General refactoring and enhancement for PoPS 1.1 release

malawrim:
Addressing issues #73, #74, and #29. Move general typedefs/enum to utils.hpp and added function to date.hpp.

### DIFF
--- a/include/pops/date.hpp
+++ b/include/pops/date.hpp
@@ -449,7 +449,7 @@ int Date::weeks_from_date(Date start)
  * \param Date, current date
  * \return String, current date as string
  */
-std::string Date::print_current_date()
+std::string Date::current_date_to_string()
 {
     std::string date = std::to_string(year_) + "-" + std::to_string(month_) + "-"
                        + std::to_string(day_);
--- a/include/pops/date.hpp
+++ b/include/pops/date.hpp
@@ -449,10 +449,21 @@ int Date::weeks_from_date(Date start)
  * \param Date, current date
  * \return String, current date as string
  */
-std::string Date::current_date_to_string()
+std::string Date::to_string()
 {
-    std::string date = std::to_string(year_) + "-" + std::to_string(month_) + "-"
-                       + std::to_string(day_);
+    std::string date = std::to_string(year_) + "-";
+    if (month_ < 10) {
+        std::string date = std::string date + "0" + std::to_string(month_) + "-";
+    }
+    else {
+        std::string date = std::string date + std::to_string(month_) + "-";
+    }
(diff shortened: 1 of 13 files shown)


malawrim:
I did not fully understand issue #29. I added a new function that returns a string of the current date, but I may have misinterpreted the issue. 

For issue #73, I was not sure about what functions to move. I don't see any that do not seem specific to the file they are in already. Were there specific ones in mind to move?

ChrisJones687:
Could move https://github.com/ncsu-landscape-dynamics/rpops/blob/ec41f5bf0f5ed88b0d74758a782b478102b92f98/src/pops.cpp#L38 to https://github.com/ncsu-landscape-dynamics/rpops/blob/ec41f5bf0f5ed88b0d74758a782b478102b92f98/src/pops.cpp#L98

to the utils.hpp as well. Then when we update rpops once these are part of the next core release we can remove them from rpops.

ChrisJones687:
@petrasovaa is this what you had in mind with #29?

petrasovaa:
> @petrasovaa is this what you had in mind with #29?

Yes, I would just shortened it to 'to_string'. It looks like the implementation would give you "2000-1-1" instead of "2000-01-01"? The latter would be preferable.

include/pops/date.hpp
```diff
@@ -443,6 +444,29 @@ int Date::weeks_from_date(Date start)
...
+    std::string date = std::to_string(year_) + "-";
+    if (month_ < 10) {
+        date = date + "0" + std::to_string(month_) + "-";
+    }
+    else {
+        date = date + std::to_string(month_) + "-";
+    }
+    if (day_ < 10) {
+        date = date + "0" + std::to_string(day_);
+    }
```
wenzeslaus, line 463: The current C++ solution for these would be a string stream, but that's actually less readable for non-C++ programmers than this, so just note in the code that this is to be replaced by the C++20 format function in the future.

include/pops/date.hpp
```diff
@@ -443,6 +444,29 @@ int Date::weeks_from_date(Date start)
...
+    else {
+        date = date + std::to_string(day_);
```
wenzeslaus, line 465: The usual style would be to use `date += std::to_string(day_);` here which documents the intention (adding to date as opposed to replacing value of date).

petrasovaa (approved):
Can't comment on template stuff but the rest looks good.

ChrisJones687 (approved):
Looks good to me. Not sure if the templating from rpops for converting from tuple to vector is the best C++ practice but it works.

PR #124, opened 2020-12-19
Fix 123

ChrisJones687:
Fixes for dispersal kernels

### DIFF
--- a/include/pops/lognormal_kernel.hpp
+++ b/include/pops/lognormal_kernel.hpp
@@ -73,7 +73,7 @@ public:
             return 0;
         }
         return (1 / (x * sigma * sqrt(2 * M_PI)))
-               * exp((pow(log(x), 2)) / (2 * pow(sigma, 2)));
+               * exp(-(pow(log(x), 2)) / (2 * pow(sigma, 2)));
     }
 
     /*!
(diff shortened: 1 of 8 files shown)


malawrim:
>

malawrim:
Sorry I meant to close a comment and closed the PR.

include/pops/lognormal_kernel.hpp
```diff
@@ -89,7 +89,7 @@ class LogNormalKernel
         if (x <= 0 || x >= 1) {
             throw std::invalid_argument("icdf: x must be between 0.0 and 1.0");
         }
-        (1 / 2) * erfc(-(log(x) / (sqrt(2) * sigma)))
+        (1 / 2) * erfc(-(log(x) / (sqrt(2) * sigma)));
```
malawrim, line 92: @ChrisJones687 could you explain the change from using inverse cumulative distribution function (icdf/quantile) to cdf?

malawrim:
Gamma kernel icdf was implemented using newtons method because there is no closed form solution. But, I had a bug in the code so if it got too far off from the right value sometimes it would continue to guess worse and worse values. I fixed this by adding a check for this that would change the direction of the next guess. So, gamma appears to be working now and I tried many different alpha and theta values and gamma was able to guess a correct answer +- 0.001.

For exponential power, I fixed one error in the formula and tested different values of beta and alpha. According to the distribution, if beta is about ~8 all distribution will be between -1 and 1. Also, the alpha value spreads the distribution, so alpha = 1 produces the distribution below and larger alpha values create long tails while smaller alpha values create short tails. It appears that the code creates the correct distribution and spread outside of the center cell should occur with beta less than 8 and more uniform spread occurs with larger alpha values.
image

For log normal, I used the original icdf formula so that gamma would work correctly. But, I fixed the problem with higher sigma values by using a different approximation for the inverse error function and now the distribution looks correct and works for higher sigma values. The higher the sigma values the more spread out the distribution becomes (peak drops and tails become longer) so with larger sigma values the spread should be more uniform and widely dispersed.

image

ChrisJones687:
@malawrim this all looks great. I made one small change in the radial kernel to account for the change you made in the power-law kernel variable order. Everything is working in the tests in R for me, except the lognormal kernel is giving a bad_array length error and the exponential power kernel isn't dispersing outside of the original cell.

The lognormal kernel doesn't give the error if we return a different value from the ICDF (i just commented out the return and had it return 2 and it works as expected for stochastic number generation. I am not sure if somewhere in the code there is a namespace issue or another problem. Can you look into this and let me know what you find out.

ChrisJones687:
Merging as it fixes 3 of the kernels. Leaving the issue open for the other 2 still.

PR #135, opened 2021-03-29
Use scaled natural kernel for overpopulation

wenzeslaus:
Instead of (mis)using antropogenic kernel as a long distance kernel for pest overpopulation movement, scale (multiply) the scale (distance) parameter of natural kernel using a coefficient.

This adds one new config variable to modify the scale parameter of radial and deterministic kernels. It is meant as modification so it is called coefficient [of scale] (`leaving_scale_coefficient`) and it defaults to 1. "Coefficient of scale" is not ideal, but "scale of scale" would not be great either.

Additionally, this moves most of the kernel creation to separate functions. Although the duplication between the functions would be unnecessary difficult to address, the similarities between the three kernels (natural, anthro, overflow) are now more clear than in the original intermingled code and easier to keep in sync if needed. The kernels are now created only when needed since they are really not only local to the run step function, but needed only for one function call.

The new functions are protected and the existing attributes are now protected instead of private because for practical reasons, similarly to other places, we consider inheritance a case when user of a class wants to deal with internals of the object.

Finally, when config is not properly used, the new behavior is that it now consistently fails while before it was sometimes passing. In the config, we now set more defaults to ensure consistent results (that is throwing an exception when not set because the default and wrong 0 is used).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ this repository.
 
 ### Changed
 - Model class internal attributes and functions are now protected instead of private
-  to allow derived classes access them to for greater flexibility.
+  to allow derived classes to access them for greater flexibility.
 - Model class kernels are now created and returned by protected functions.
 - Config class has now more defaults and subsequent setup now consistently fails when
   required values were not set.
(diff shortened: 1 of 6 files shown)


ChrisJones687 (approved):
I like the update to the kernel creation step. Great Job!

wenzeslaus:
Thanks!

PR #136, opened 2021-05-03
added total_hosts to model

ChrisJones687:
Addresses #128 and part of #87


wenzeslaus (approved):
Great. It even fixes the issues with `total_hosts = [I] + S` in tests I mentioned.

PR #138, opened 2021-05-11
Flexible Mortality

ChrisJones687:


ChrisJones687:
closes #137

include/pops/simulation.hpp
```diff
@@ -190,38 +190,57 @@ class Simulation
...
+                    if (index == 0) {
+                        mortality_in_index = mortality_tracker_vector[index](i, j);
+                    }
+                    else {
+                        mortality_in_index =
+                            mortality_rate * mortality_tracker_vector[index](i, j);
+                    }
+                    mortality_tracker_vector[index](i, j) -= mortality_in_index;
+                    died(i, j) += mortality_in_index;
+                    mortality_current_step += mortality_in_index;
```
wenzeslaus, line 232: The value seems to be unused like with `mortality_current_year`. Probably left over from debugging or from the idea that there would be two "died" handled by this function (one for year and one total). I suggest to remove the variable.

include/pops/simulation.hpp
```diff
@@ -190,38 +190,57 @@ class Simulation
         }
     }
 
+    /** kills infected hosts based on mortality rate and timing
+     *
```
wenzeslaus, line 194: Please add a paragraph about mortality_rate and mortality_tracker_vector, like ...its size needs to be at least `mortality_time_lag + 1` ... first item is the longest infected ... and all hosts in it die regardless of mortality_rate.

include/pops/simulation.hpp
```diff
@@ -190,38 +190,57 @@ class Simulation
...
+        int max_index = mortality_tracker_vector.size() - mortality_time_lag - 1;
+
+        for (auto indices : suitable_cells) {
+            int i = indices[0];
+            int j = indices[1];
+            for (int index = 0; index <= max_index; index++) {
+                int mortality_in_index = 0;
+                if (mortality_tracker_vector[index](i, j) > 0) {
+                    if (index == 0) {
+                        mortality_in_index = mortality_tracker_vector[index](i, j);
```
wenzeslaus, line 224: Note here that this is to ensure removal/killing of the remaining infected hosts in the the last year of tracking mortality.

tests/test_model.cpp
```diff
@@ -447,15 +456,18 @@ int test_model_sei_deterministic()
...
+    config.mortality_frequency = "year";
+    config.mortality_frequency_n = 1;
     config.create_schedules();
 
     config.deterministic = true;
 
     std::vector<std::vector<int>> movements;
 
-    unsigned num_mortality_years = config.num_mortality_years();
+    unsigned num_mortality_steps = 1;
```
wenzeslaus, line 468: Is `config.num_mortality_years()` giving something else than 1? If yes, please note it here, we need to investigate or at least document it better. Or given you have `config.use_mortality = false;` above, shouldn't this be 0. That should work as mortality should not be touched with `use_mortality = false`. Both of these tests should work with `use_mortality = false` which is already set as the default in config. Maybe this is poorly defined in the API or impossible to do at this point because here we need 0 steps for something which optionally does not have a schedule. In any case, a lengthy but right and safe idiom would be:

```C++
unsigned num_mortality_steps = 0;
if (config.use_mortality)
    num_mortality_steps = config.num_mortality_steps();
```

...or in this case without the `if` part, only with a comment about not using mortality.

I think these two tests didn't have to care until now which would be nice to preserve, but perhaps with the above better expressed in the code.

tests/test_mortality.cpp
```diff
@@ -42,28 +42,57 @@ using std::endl;
...
     int ew_res = 30;
     int ns_res = 30;
     double mortality_rate = 0.50;
-    int current_year = 2018;
-    int first_mortality_year = 2018;
+    int mortality_time_lag = 0;
     std::vector<std::vector<int>> suitable_cells = {{0, 0}, {0, 1}, {1, 0}, {1, 1}};
-    Simulation<Raster<int>, Raster<double>> simulation(42, infected, ew_res, ns_res);
+    Simulation<Raster<int>, Raster<double>> simulation(
+        42, infected.rows(), infected.cols());
```
wenzeslaus, line 60: Simulation constructor really takes resolution, not raster size. (Suggest a different naming if needed. These are short but perhaps impossible to parse out.)

The resolution is used for dispersal, so not in this test, so it could be even 0, anything as long as it is not misleading.

include/pops/config.hpp
```diff
@@ -71,8 +71,10 @@ class Config
     bool use_treatments{false};
     // Mortality
     bool use_mortality{false};
+    std::string mortality_frequency;
+    unsigned mortality_frequency_n;
     double mortality_rate{0};
-    int first_mortality_year{0};  // TODO: document that it starts at 1, not 0
+    int mortality_time_lag{0};  // TODO: document that it starts at 1, not 0
```
wenzeslaus, line 77: Since it is lag and not year, the TODO comment is no longer needed. Well, it is actually wrong, but more importantly, there is no possible confusion between indices and ordinal numbers for years.  Documentation `///< Time lag of mortality in mortality steps` would be of course nice.

wenzeslaus (changes requested):
Looks good. Simpler code! I checked the numbers in the test, too.

tests/test_mortality.cpp
ChrisJones687: The Simulation constructor takes rows and columns and config.ew_res and ns_res are set as part of the config setup.

tests/test_model.cpp
ChrisJones687: `config.num_mortality_steps()` is returning 0. Which is to be expected and correct behavior but to run the model we need mortality tracker to be of length 1 or greater. mortality_tracker is used in other modules even if `use_mortality=false` . So for the tests this should be 1 we could get rid of the num_mortality_steps = 1 and just use 1 to assign the mortality tracker.

tests/test_mortality.cpp
wenzeslaus: Wait, that's wrong. That is true for (some) kernels, but Simulation needs rows cols.

include/pops/simulation.hpp
wenzeslaus: Isn't the variable still defined above, i.e., defined but unused. No big deal, we can get it next time we improve compile warnings status.

wenzeslaus (approved):

include/pops/simulation.hpp
ChrisJones687: Good catch, I removed it.

PR #141, opened 2021-05-14
Update movements

ChrisJones687:


include/pops/simulation.hpp
```diff
@@ -247,80 +261,128 @@ class Simulation
...
-                    if (movement_stochasticity_) {
-                        std::poisson_distribution<int> distribution(infected_mean);
-                        infected_moved = distribution(generator_);
-                    }
-                    else {
-                        infected_moved = infected_mean;
-                    }
+            auto total_infecteds = infected(row_from, col_from);
+            auto suscepts = susceptible(row_from, col_from);
+            auto expose = total_exposed(row_from, col_from);
```
wenzeslaus, line 301: I don't remember whether you actually want to add total_exposed in general or it is just for this. If you put the `for (auto& raster : exposed) {` loop into forming of categories starting the index at 4, you won't need total_exposed. You would not have (total_)infected here in categories, but you would loop through exposed after the draw to get the (total_)infected. However, if you want total_exposed, then the current code is actually better then what I'm suggesting.

include/pops/simulation.hpp
```diff
@@ -42,6 +42,19 @@ void rotate_left_by_one(Container& container)
...
+ */
+template<typename Generator>
+std::vector<int> draw_n_from_v(std::vector<int> v, unsigned n, Generator& generator)
+{
+    if (n > v.size())
+        n = v.size();
+
+    std::shuffle(v.begin(), v.end(), generator);
+    v.erase(v.begin() + n, v.end());
+    return (v);
```
wenzeslaus, line 55:
```suggestion
    return v;
```

include/pops/simulation.hpp
```diff
@@ -42,6 +42,19 @@ void rotate_left_by_one(Container& container)
     std::rotate(container.begin(), container.begin() + 1, container.end());
 }
 
+/** Draws n elements from a vector. Expects n to be less than v.size().
```
wenzeslaus, line 45:
```suggestion
/** Draws n elements from a vector. Expects n to be equal or less than v.size().
```

wenzeslaus (changes requested):
I have some mostly C++ issues and I have a question about total_exposed.

include/pops/simulation.hpp
ChrisJones687: Other than using it here, it isn't useful for the model at this point but could be very useful for visualization purposes on the dashboard. I was planning on calculating it in the interface but now it can just be exported saving multiple loops in the interface. A long way to say that I think it is worth keeping

include/pops/simulation.hpp
```diff
@@ -247,80 +261,128 @@ class Simulation
...
+            // check that the location with host movement is in suitable cells.
+            // Since suitable-cells comes from the total hosts originally. The
+            // the first check is for total_hosts
+            if (total_hosts(row_to, col_to) == 0) {
+                int match = 0;
+                for (auto indices : suitable_cells) {
+                    int i = indices[0];
+                    int j = indices[1];
+                    if ((i == row_to) & (j == col_to)) {
+                        match += 1;
```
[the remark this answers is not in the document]
ChrisJones687: I moved the push back into the loop.

include/pops/simulation.hpp
wenzeslaus: Sounds good!

wenzeslaus (approved):

PR #143, opened 2021-06-03
Fix 139

ChrisJones687:


wenzeslaus:
Can you update the change log too maybe mentioning both the fix and the new citation since we want the citation to get into the release description.

include/pops/treatments.hpp
```diff
@@ -175,29 +176,40 @@ class SimpleTreatment : public BaseTreatment<IntegerRaster, FloatRaster>
...
+        IntegerRaster& resistant,
+        IntegerRaster& total_hosts,
         const std::vector<std::vector<int>>& suitable_cells) override
     {
         for (auto indices : suitable_cells) {
             int i = indices[0];
             int j = indices[1];
+            int exposed_changed = 0;
+            int infected_changed = 0;
+            int susceptible_changed = 0;
```
wenzeslaus, line 188: Please name these something like `new_infected, new_susceptible`, changed is little misleading. For `exposed_changed`, maybe `new_exposed_total` to be really clear.

include/pops/treatments.hpp
```diff
@@ -175,29 +176,40 @@ class SimpleTreatment : public BaseTreatment<IntegerRaster, FloatRaster>
...
             }
             else if (this->application_ == TreatmentApplication::AllInfectedInCell) {
-                infected(i, j) = this->map_(i, j) ? 0 : infected(i, j);
+                infected_changed = this->map_(i, j) ? 0 : infected(i, j);
+                infected(i, j) = infected_changed;
             }
             for (auto& raster : exposed) {
                 if (this->application_ == TreatmentApplication::Ratio) {
                     raster(i, j) = raster(i, j) - (raster(i, j) * this->map_(i, j));
+                    exposed_changed += raster(i, j) - (raster(i, j) * this->map_(i, j));
```
wenzeslaus, line 200: Here you need a new local variable, something like `new_exposed_individual` or rather a simple `new_value`, and 1) assign it to `raster(i, j)` and add it to the new exposed total.

include/pops/treatments.hpp
```diff
@@ -244,6 +256,7 @@ class PesticideTreatment : public BaseTreatment<IntegerRaster, FloatRaster>
         std::vector<IntegerRaster>& exposed_vector,
         IntegerRaster& susceptible,
         IntegerRaster& resistant,
+        IntegerRaster&,
         const std::vector<std::vector<int>>& suitable_cells) override
     {
```
wenzeslaus, line 261: We newly have `UNUSED` defined in `utils.hpp`, so you can do:

```cpp
#include "utils.hpp"
...
    IntegerRaster& total_hosts,
...
{
    UNUSED(total_hosts);
```

which makes it clearly documented at least in the code.

tests/test_treatments.cpp
```diff
@@ -67,19 +71,22 @@ int test_application_all_inf()
...
 
     treatments.add_treatment(
         tr1, Date(2020, 1, 1), 0, TreatmentApplication::AllInfectedInCell);
-    treatments.manage(0, infected, exposed, susceptible, resistant, suitable_cells);
+    treatments.manage(
+        0, infected, exposed, susceptible, resistant, total_hosts, suitable_cells);
 
     Raster<int> treated = {{0, 3}, {5, 42}};
     Raster<int> inf_treated = {{0, 0}, {0, 40}};
+    auto th_treated = treated + inf_treated + resistant;
```
wenzeslaus, line 86: Unused? `total_hosts == th_treated`?

README.md
```diff
@@ -20,20 +20,12 @@ required version.
...
-
-* Tonini, Francesco, Douglas Shoemaker, Anna Petrasova, Brendan Harmon,
-  Vaclav Petras, Richard C. Cobb, Helena Mitasova,
-  and Ross K. Meentemeyer, 2017.
-  Tangible geospatial modeling for collaborative solutions
-  to invasive species management.
-  *Environmental Modelling & Software* 92: 176-188.
-  [DOI: 10.1016/j.envsoft.2017.02.020](https://doi.org/10.1016/j.envsoft.2017.02.020)
+* Jones, Chris, Jones, Shannon, Petrasova, Anna, Petras, Vaclav, Gaydos, Devon, 
+  Skrip, Megan, Yu, Takeuchi, Bigsby, Kevin, and Meentemeyer, Ross, 2021.
```
wenzeslaus, line 24: This should probably be "Jones, Chris, Shannon Jones, Anna Petrasova, ..." or "Jones, C., Jones, S., Petrasova, A., ..." but if you don't have it at hand, don't worry about that, I can change it when I do some bigger updates for DOI in citation file etc.

wenzeslaus (changes requested):

include/pops/treatments.hpp
```diff
@@ -175,29 +177,43 @@ class SimpleTreatment : public BaseTreatment<IntegerRaster, FloatRaster>
...
-                    raster(i, j) = raster(i, j) - (raster(i, j) * this->map_(i, j));
+                    new_exposed_individual =
+                        raster(i, j) - (raster(i, j) * this->map_(i, j));
+                    raster(i, j) = new_exposed_individual;
+                    new_exposed_total += new_exposed_individual;
                 }
                 else if (
                     this->application_ == TreatmentApplication::AllInfectedInCell) {
                     raster(i, j) = this->map_(i, j) ? 0 : raster(i, j);
+                    new_exposed_total += this->map_(i, j) ? 0 : raster(i, j);
```
wenzeslaus, line 209: Can you please do one more change and do this through assigning new_exposed_individual? I just spend some time doubting whether this is even correct. It is correct because `raster(i, j)` is either the original value or 0, but it would be wrong if it would be, e.g., multiplied. Anyway, for clarity and maybe speed, it still is worth to change I think.

wenzeslaus (changes requested):

include/pops/treatments.hpp
ChrisJones687: good catch I missed that in the else if statement.

PR #145, opened 2021-07-16
Add network kernel

### LINKED ISSUE #15: Add Transportation Network Model
Based on the spotted lanternfly case study it is clear the need for some type of spread via transportation networks. For example, spotted lanternfly spreads long distances along railways by hitchhiking on trains and hopping off along the route. This could be done via a cost matrix (railway corridors speed up spread, forest areas slow spread, etc.

wenzeslaus:
This adds a network kernel type for travel dispersers across an spatially-explicit network such a railroad network.  

Kernel:

- New network kernel type which behaves exactly as any other kernel taking source row and column and returning destination row and column.
- The network kernel is available as natural and anthorpogenic, but it is meant to be used as anthorpogenic.
- Kernel interface now includes test for eligibility of a kernel to handle given source cell. This is to ensure that when network kernel gets row and column, there is actually a network node to travel from. If there is no node, an exception is thrown. The random challenge for anthorpogenic kernel now happens only when the anthorpogenic can handle the given row and column because witching back to natural kernel after already passing the random challenge for anthorpogenic would drastically change the simulated proportion of natural versus anthorpogenic spread. Practically, when the main kernel decides for an anthorpogenic kernel, there is no way back and the decision is made in dispersing code which doesn't know any details about kernels. (Eligibility check is not implemented for all kernels, but only where needed.)
- The kernel class itself is minimal taking network object in the constructor and delegating eligibility and actual travel to it.
- Network is required as part of the Model inputs. A convenience function for creation of null (empty) network is provided (useful mostly for tests).

Functions and algorithm:

- The network is consists of nodes connected by edges. Edges are spatially represented as segments. Edges themselves don't carry cost and have meaning only as indicators of existence of a connection between nodes. Cost is travel time determined by advancing through the segment.
- The input network is a list of nodes and their coordinates and a list of edges and the corresponding segments. The input is used as-is. The network is agnostic towards how the segments look like in terms of crossing each other or other nodes.
- The network representation is combination of STL containers rather than a graph, i.e., edges and their segments are mappings rather than graph structures. 
- Nodes are the hop-on locations for dispersers. The dispersers can hop-off anywhere.
- All XY coordinates of nodes and segment points are converted to row and column using bounding box and resolution.
- Coordinates for nodes and segment are treated separately, so input segment coordinates should include the node coordinates.
- Segments and edges are treated together in the input. Edges are then stored separately in both directions as connected (neighboring) nodes to each node. Segments are stored only in the original direction and reversed as needed.
- Standalone nodes (without any connection) are allowed and handled as no movement from the source cell.
- Only edges (segments) with both nodes in the bounding box are considered, i.e., input network is reduced to the bounding box during reading, but clipping happens on the level of whole edges, not in the middle of a segment.

Speed:

- Parsing input takes significant amount of time more visible for large networks and small amount of network travel (e.g., rare network spread events). However, most time is now spend in places which cannot be further optimized without change to binary format such as string to double conversions.
- Network format is custom text format resembling CSV geared towards parsing, not readability.
- If the input network match segment points match the resolution, the reading is faster, but extra points are not converted to cells and the resulting network matches the resolution.
- Only non-const (write) functions for network are those loading the network, so an existing network can be shared by multiple kernel or model instances.
- The node-to-node matrix is sparse (esp. with indexing by node ids twice larger than their count), so using std::map, not a matrix- like object (comparing to Raster, speedup was significant (from >1s to 0.1s) and a lot of memory saved.
- All needed mappings are constructed ahead of time when reading the network to be available exactly as they are needed by the algorithm, i.e., they are precomputed. So, functions generally return const references to existing containers.
- Iteration over segments happens in both directions, so a view which abstracts direction is constructed in segment getter to avoid need to return reversed container and a copy. (This reduces overall malloc and free calls to a half while still keeping the segment lookup and direction issues out of the main travel function and storing only the one direction present in the input.)
- The cost of the look-ups in mappings in STL containers versus using custom graph structures was not tested.

Tests:

- A simple test of the kernel used in the model is included.
- Main tests of network are directly testing the Network class.
- Network class has several functions specifically for tests and used only by tests. However, no test-related functionality is included in the functions used for the kernel even if that means that the testing is less convenient. 
- The test program can take parameters and perform several tasks to support profiling, manual testing, and examination of the internal network.
- The tests are using only topologically clean input network (messy data is not expected to work).

Resolves #15.


wenzeslaus:
I merged it as discussed during our meeting. Now more modifications may or should follow including:

- Evaluate need for the nodes file (node IDs and coordinates _are_ included in the segments file).
- Overall update and fix of the documentation including adding some of the details from the commit message to the comments in the code.
- Support for multiple networks (may require more dynamic kernels or at least dynamic kernel for networks).
- Support for configuration files (some code already included, updating minimal C++ version may help).
- Removal of the network from natural kernel (introducing another switch-like kernel might be enough).

PR #147, opened 2021-08-06
Support time_step as frequency in scheduling

wenzeslaus:
This tries to remove the need for [this code](https://github.com/ncsu-landscape-dynamics/rpops/blob/62d919121d84ed33782ef58cffaf87f16e4d7ca8/src/pops.cpp#L133-L135) in rpops:

```cpp
    if (output_frequency == "time_step") {
        output_frequency = time_step;
    }
```

where time_step is 'day', 'week', 'month', ...

Any simulation unit + n combination will now work with `output_frequency == "time_step"`. This is not used with rpops if I recall correctly, but the code there seems to allow that (?).

The logic is that we know what is the step in the scheduler, so scheduling every step is just schedule every nth step with n = 1. With this in place, the rpops code does not have to get the step from parameters.

Given that we have `every_n_steps` this new option could be `every_step` not `time_step`, but perhaps it can be both since for other options, we support multiple values (`day` and `daily`). `time_step` seems to be a valid description context where n != 1, right?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ the tagged revision was already tagged before.
 - Established dispersers from a cell are now tracked. (Chris Jones)
 - Collection of general-use function for container handling for internal use.
 - Template parameter for random number generator to Model and Simulation.
+- New frequency values, every_step and time_step, can be used to schedule
+  action every step of the simulation. (Vaclav Petras)
+  * This was implemented in rpops for simple cases. The new version in Core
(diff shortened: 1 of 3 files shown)


ChrisJones687 (approved):

ChrisJones687:
Yes if time_step is set in the R interface. Internally it changes to for example output_frequency = time_step. I still need to add a time_step_n parameter so that the time step can be week with n = 2.

PR #156, opened 2021-11-03
Remove node file from network

wenzeslaus:
This removes the need to load the network as a file with nodes and another file with segments. Now, only the file with segments is needed since it anyway includes nodes and their coordinates.

There is a slight speed up, but only very small because the node file was small comparing to segments file and the more complex data structures needs to be filled anyway.

However, this simplifies the inputs and required data.

### DIFF
--- a/include/pops/network.hpp
+++ b/include/pops/network.hpp
@@ -182,13 +182,14 @@ public:
     /**
      * @brief Load nodes and segments from input streams.
      *
-     * @param node_stream Input stream with records for nodes
-     * @param segment_stream Input stream with records for segments
+     * @param stream Input stream with records for network segments
      * @param allow_empty True if the loaded network can be empty (no nodes)
      *
-     * The nodes stream is a CSV with rows `node_id,X,Y`.
-     * Segment stream is custom text format resembling CSV geared towards parsing, not
-     * readability, with rows `node_id_1,node_id_2,X1;Y1;X2;Y2;X3;Y3;...`.
+     * Network segment stream is custom text format resembling CSV geared towards
+     * parsing, not readability, with rows `node_id_1,node_id_2,X1;Y1;X2;Y2;X3;Y3;...`.
+     * Nodes, edges, and segment spatial representation are read from the network
+     * segement stream. Node coordinates are taken from the first and last coordinate
+     * pair in the segment.
      *
      * The function can take any input stream which behaves like std::ifstream or
      * std::istringstream. These are also the two expected streams to be used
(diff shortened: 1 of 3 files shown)


include/pops/network.hpp
```diff
@@ -552,19 +534,25 @@ class Network
                 if (segment.empty() || segment.back() != new_point)
                     segment.emplace_back(new_point);
             }
+            // If either end nodes of the segment is not in the extent, skip it.
```
petrasovaa, line 537:
```suggestion
            // If either node of the segment is not in the extent, skip the segment.
```

include/pops/network.hpp
```diff
@@ -144,43 +171,42 @@ class Network
...
-     * std::ifstream node_stream{node_filename};
-     * std::ifstream segment_stream{segment_filename};
-     * network.load(node_stream, segment_stream);
+     * std::string filename = "network.txt"
+     * std::ifstream stream{filename};
+     * network.load(stream);
      * ```
      *
-     * The input network is a list of nodes and their coordinates and a list of edges
+     * The input network a list of edges which are pairs of nodes
```
petrasovaa, line 179:
```suggestion
     * The input network is a list of edges which are pairs of nodes
```
Should it rather be "list of segments which are pairs of nodes and the corresponding edges"?
Or "where each segments consists of a pair of nodes and a corresponding edge".

include/pops/network.hpp
```diff
@@ -120,21 +124,44 @@ class Network
...
      * @return True if XY is in the bounding box, false otherwise.
      */
-    bool out_of_bbox(double x, double y) const
+    bool xy_out_of_bbox(double x, double y) const
     {
         return x > bbox_.east || x < bbox_.west || y > bbox_.north || y < bbox_.south;
     }
 
+    /**
+     * @brief Test if a cells is in the bounding box.
```
petrasovaa, line 133:
```suggestion
     * @brief Test if a cell is in the bounding box.
```

include/pops/network.hpp
```diff
@@ -120,21 +124,44 @@ class Network
...
+     * @param col Column in the raster grid
+     * @return True if cells is in the bounding box, false otherwise.
+     */
+    bool row_col_out_of_bbox(RasterIndex row, RasterIndex col) const
+    {
+        return row > max_row_ || row < 0 || col > max_col_ || col < 0;
+    }
+
+    /**
+     * @brief Test if a cells is in the bounding box.
```
petrasovaa, line 144:
```suggestion
     * @brief Test if a cell is in the bounding box.
```

include/pops/network.hpp
```diff
@@ -120,21 +124,44 @@ class Network
...
+     * @param cell Pair consisting of a row and column indices
+     * @return True if cells is in the bounding box, false otherwise.
```
petrasovaa, line 146:
```suggestion
     * @return True if cell is in the bounding box, false otherwise.
```

petrasovaa (approved):
Some minor typos/language considerations.

include/pops/network.hpp
```diff
@@ -172,43 +199,42 @@ class Network
...
-     * std::ifstream segment_stream{segment_filename};
-     * network.load(node_stream, segment_stream);
+     * std::string filename = "network.txt"
+     * std::ifstream stream{filename};
+     * network.load(stream);
      * ```
      *
-     * The input network is a list of nodes and their coordinates and a list of edges
+     * The input network is a list of edges which are pairs of nodes
      * and the corresponding segments. The input is used as-is. The network is agnostic
```
wenzeslaus, line 208: @petrasovaa said:

> Should it rather be "list of segments which are pairs of nodes and the corresponding edges"?
Or "where each segments consists of a pair of nodes and a corresponding edge".

wenzeslaus: Addressed in 5963d739fae0e74a9b63403605e8bbeaac97d3af

PR #161, opened 2021-11-19
Replace time and speed by distance

wenzeslaus:
* Replace time by distance.
* Remove speed.
* Network test with exact destination cells stays the same because cost per cell is still 1 in that test.


petrasovaa (approved):
The idea behind the speed was to model how the pest can spread more easily when a train is slow or standing in a station. If we ever have some train data, it may be interesting to reintroduce the concept there.

PR #166, opened 2022-01-04
Snap points to the closest node in network

wenzeslaus:
Optionally, instead of ending in somewhere in the segment, the network travel ends at the start or end node
of a segment (whichever is the closest).

To support the snapping based on distance, each segment now has explicit total cost (distance), so Segment and SegmentView
need to be classes, not just using.

Having the total cost (distance) for segment available allows the travel function to not iterate over a segment when its cost
is lower than the remaining distance (cost) to travel (spend).

This contains only the changes in the network itself.


include/pops/network.hpp
```diff
@@ -544,6 +664,68 @@ class Network
...
+        std::string label;
+        int column_number = 0;
+        while (std::getline(line_stream, label, delimeter)) {
+            column_number++;
+            if (column_number == 1 && label != "node_1") {
+                // The right label is not there. Assuming it is without a header.
+                stream.seekg(starting_position);
+                break;
+            }
+            if (label == "cost") {
```
petrasovaa, line 718: Shouldn't the header be documented somewhere?

petrasovaa (approved):

ChrisJones687 (approved):

include/pops/network.hpp
wenzeslaus: Documented in the load method description where the rest of the format is documented.

PR #169, opened 2022-01-24
Add missing throw statement

wenzeslaus:
The throw keyword was missing and the statement had no effect.

### DIFF
--- a/include/pops/network.hpp
+++ b/include/pops/network.hpp
@@ -564,12 +564,12 @@ protected:
             auto node_1_id = node_id_from_text(node_1_text);
             auto node_2_id = node_id_from_text(node_2_text);
             if (node_1_id < 1 || node_2_id < 1) {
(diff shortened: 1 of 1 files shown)


wenzeslaus:
This PR merged in January 2022 fixes a check for edges which begin and end with the same node (loops without additional nodes). A file I tested from our GDrive named `US_primary_secondary_roads_segments.csv` works with the previous version, but fails with this PR for edge "node 1: 118, node 2: 118" (line 123): `118,118,-9627404.15629989;3847220.97315297;-9626902.3280354;3847282.4426179;-9627299.29333956;3847599.84164896;-9627404.15629989;3847220.97315297`. This is a good thing if we still stand behind the idea that travel to the same node without visiting another node first does not make sense. However, we need fix our files (and possibly the generator).

This fix didn't make it into the v2.0.0 release (December 2021), so it was sitting on main till now.

Issue #173, opened 2022-03-10
More flexible lethal temperature


ChrisJones687:
Make the lethal temperature function more flexible by allowing it to be applied at a specified date and instead of using raw temperature values the raster that it uses contains values from 0 to 1. This allows for partial removal due to less ideal conditions during diapause that lead to lower emergence from diapause (i.e. higher over-winter mortality that isn't binary).

PR #175, opened 2022-03-15
Add option to move percentage of exposed/infected hosts to susceptibles

petrasovaa:
Attempts to address #173. It is written similarly to lethal temperature.

Removes certain percentage of exposed/infected host and makes them susceptible.
The scheduling is done in the same way as lethal temperature, so it is scheduled yearly on the first of certain month.

So far I didn't add any tests and didn't remove lethal temperature. Let me know if this addresses the issue or we want something little different.

### DIFF
--- a/include/pops/simulation.hpp
+++ b/include/pops/simulation.hpp
@@ -230,13 +230,15 @@ public:
             if (survival_rate(i, j) < 1) {
                 int removed = 0;
                 // remove percentage of infestation/infection in the infected class
-                int removed_infected = infected(i, j) - infected(i, j) * survival_rate(i, j);
+                int removed_infected =
+                    infected(i, j) - infected(i, j) * survival_rate(i, j);
                 infected(i, j) -= removed_infected;
                 removed += removed_infected;
                 // remove the same percentage in each exposed cohort
                 int total_removed_exposed = 0;
                 for (auto& raster : exposed) {
-                    int removed_exposed = raster(i, j) - raster(i, j) * survival_rate(i, j);
+                    int removed_exposed =
+                        raster(i, j) - raster(i, j) * survival_rate(i, j);
                     raster(i, j) -= removed_exposed;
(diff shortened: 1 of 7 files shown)


ChrisJones687:
This is basically what I had in mind. I have two questions:
1. should we handle mortality in this. I know we don't in lethal temp (maybe we should there). The logic for not was that typically pathogens don't completely dieoff but don't reproduce during harsh conditions. Here thinking about if we model grapes as an SLF host they could experience mortality.
2. would it make sense to make it take in both month and date. Could be useful for weekly timesteps. Although removing anytime before spread starts would give the same answer computationally, not sure if it would be preferred for realism to have a more precise date.

ChrisJones687 (approved):
Looks great!

include/pops/simulation.hpp
```diff
@@ -347,43 +433,28 @@ class Simulation
...
-                std::vector<int> mortality_draw =
-                    draw_n_from_v(mortality_categories, infected_moved, generator_);
-                index = 0;
-                for (auto& raster : mortality_tracker_vector) {
-                    auto mortality_moved_in_cohort =
-                        std::count(mortality_draw.begin(), mortality_draw.end(), index);
-                    raster(row_from, col_from) -= mortality_moved_in_cohort;
-                    raster(row_to, col_to) += mortality_moved_in_cohort;
+                    std::cout << mortality_draw[index] << std::endl;
+                    std::cout << raster(row_from, col_from) << std::endl;
```
wenzeslaus, line 455: Forgotten debug output.

wenzeslaus (changes requested):
Both code and test look good! I found only two forgotten debug lines.

include/pops/simulation.hpp
petrasovaa, line 455:
```suggestion
```

wenzeslaus (approved):

PR #181, opened 2022-04-19
Rename network movement types

### LINKED ISSUE #180: New names for network types
Config for the network is now `std::string network_type` with values `"travel"`, `"snap"`, and `"step"`.

The three options are:
1. Go through the network pixel by pixel using the cost (distance), stop at any pixel.
2. Go through the network pixel by pixel using the cost (distance), stop at nodes.
3. Go through the network from node to node. 

There is actually not much difference between 2 and 3 spatially or in the implementation, but the big difference is that edge cost is considered in 2 and not in 3.

The option 3 is different from 1 and 2 in use of probability for edges (picking which edge to go from a node). This is now completely missing in 1 and 2, but it can be added. At the same time, unique edge probabilities are optional for 3 and the use of the unique edge probabilities (as opposed to equal probabilities) is determined by the probability column being present in the network file rather than by the configuration, so 3 is not defined by use of edge probabilities.

These are options I came up with:
* travel, snap, step (current, step and snap are too similar)
* travel, snap, jump (jump is the only other thing I came up with)
* walk, snap, jump (walk seems to pair well with jump)
* cost, cost-snap, jump (jump with cost would be like cost with snapping ignoring the edge probabilities difference)
* pixel, node, jump (getting to the interaction with the physical world, but node and jump actually interact in the same way)
* cell, node, jump (PoPS Core mostly uses cell, not pixel)
* spatial, nodes, jump (the spatial part is what makes it different from classic networks in network modeling or GIS)
* open carriage, closed carriage, teleport (based on the mode of transportation, but trying to be abstract)
* open carriage, closed carriage, transporter (Star Trek)
* open carriage, closed carriage, wormhole (Stargate)
* open carriage, closed carriage, plane (less science fiction, but the two words per option are sub-optimal)
* chariot, wagon, transporter (too far?)

_Originally posted by @wenzeslaus in https://github.com/ncsu-landscape-dynamics/pops-core/issues/168#issuecomment-1098150053_

wenzeslaus:
- Renames travel to walk.
- Renames snap to jump.
- Renames step to teleport.
- Renames network_type to network_movement.

Closes #180.

### DIFF
--- a/include/pops/anthropogenic_kernel.hpp
+++ b/include/pops/anthropogenic_kernel.hpp
@@ -59,9 +59,9 @@ std::unique_ptr<KernelInterface<Generator>> create_anthro_kernel(
     else if (anthro_kernel == DispersalKernelType::Network) {
         using Kernel =
             DynamicWrapperKernel<NetworkDispersalKernel<RasterIndex>, Generator>;
-        if (config.network_type == "step")
+        if (config.network_movement == "step")
             return std::unique_ptr<Kernel>(new Kernel(network));
-        bool snap = config.network_type == "snap" ? true : false;
+        bool snap = config.network_movement == "snap" ? true : false;
         return std::unique_ptr<Kernel>(new Kernel(
             network, config.network_min_distance, config.network_max_distance, snap));
     }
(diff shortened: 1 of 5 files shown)


ChrisJones687 (approved):
looks great!

petrasovaa (approved):

Issue #174, opened 2022-03-11
Soil Borne Dispersal


ChrisJones687:
Add in the ability of a soil-borne disease to spread from the soil when the disease was present in the past. For example, for late blight the disease over-winters in the soil after the tomato plants are removed but can last in the soil for multiple years providing a source to infect newly planted tomato or potato plants and thus lead to new spread when soil splashes on leaves.

PR #182, opened 2022-05-03
Add soils and create a separate function to move dispersers to cell

wenzeslaus:
- Instead of run_step, data is passed using a separate function.
- Soil class is used in generate and disperse, but details are in the class.
- Disperse has now soils as a an alternative disperser target.


include/pops/simulation.hpp
```diff
@@ -120,6 +121,119 @@ inline ModelType model_type_from_string(const char* text)
...
+                "List of rasters of SpoilPool needs to have at least one item");
+        }
+    }
+
+    SoilPool() : active_{false} {}
+
+    bool active()
+    {
+        return active_;
+    }
```
wenzeslaus, line 175: Not needed. Driven by smart pointers.

include/pops/simulation.hpp
```diff
@@ -521,6 +637,9 @@ class Simulation
                 else {
                     dispersers_from_cell = lambda * infected(i, j);
                 }
+                if (soil_pool_ && soil_pool_->active())
+                    dispersers_from_cell +=
+                        soil_pool_->dispersers_from(i, j, generator_);
```
wenzeslaus, line 642: Can only move one cell, so generate separately from the other disperses and infect separately in one cell.

include/pops/simulation.hpp
```diff
@@ -607,6 +726,11 @@ class Simulation
                         outside_dispersers.emplace_back(std::make_tuple(row, col));
                         continue;
                     }
+                    if (soil_pool_
+                        && soil_pool_->active() /* && to_soil_probabiliy */) {
+                        soil_pool_->disperser_to(i, j, generator_);
+                        continue;
+                    }
```
wenzeslaus, line 733: These are taken from the general disperser pool by ratio, but then applied only to the same cell the disperser is in ("trivial kernel"). This can be done right away in generate for performance reasons.

wenzeslaus (commented):
Adding notes from VP, CJ, AP meeting as a review.

include/pops/simulation.hpp
wenzeslaus: Property removed, pointers already there.

include/pops/simulation.hpp
wenzeslaus: Disperses from soil are now retrieved in disperse (so not generated in the current terminology, but obtained from the storage).

include/pops/simulation.hpp
wenzeslaus: Soil is acting as an additional disperser storage besides the standard disperers pool, so storing now happens after generation in the generate function.

include/pops/simulation.hpp
```diff
@@ -608,45 +573,84 @@ class Simulation
...
+                        total_exposed,
+                        weather,
+                        weather_coefficient,
+                        establishment_probability);
+                }
+            }
         }
     }
 
+    // private
```
ChrisJones687, line 611: Make it clear the purpose for this and how it differs from dispers_to in soils

include/pops/simulation.hpp
```diff
@@ -901,6 +905,20 @@ class Simulation
...
         }
     }
+
+    /**
+     * @brief Activate storage of dispersers in soil
+     *
+     * @param soil_pool Soils pool object to use for storage
+     * @param dispersers_percenatge Percentage of dispersers moving to the soil
+     */
+    void activate_soils(
```
ChrisJones687, line 915: Highlight choice for this in comments

ChrisJones687:
Closes #174

PR #202, opened 2023-09-18
CI: Check header files with Cppcheck

wenzeslaus:
Checks header files with a separate suppressions file and updates versions in the workflow.


wenzeslaus:
Get in after #184.

PR #210, opened 2023-12-18
Disable the infected-to-dead transition when mortality rate is 0

wenzeslaus:
A [new test](https://github.com/ncsu-landscape-dynamics/r.pops.spread/pull/70/commits/c8d073ad86ec3fa57e048ce5eee214d33552aa0a) with enabled mortality and mortality rate 0 fails in r.pops.spread CI without this change, but locally passes with this change. The test is set in the way that mortality rate 0 should produce same values as no mortality enabled. Test is for one host only.

### DIFF
--- a/include/pops/host_pool.hpp
+++ b/include/pops/host_pool.hpp
@@ -795,6 +795,9 @@ public:
      * individuals is multiplied by the mortality rate to calculate the number of hosts
      * that die that time step.
      *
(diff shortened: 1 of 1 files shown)


ChrisJones687 (approved):

wenzeslaus:
This allows mortality to be enabled and disabled for specific hosts in a multi-host case by setting the mortality rate to zero for hosts where mortality does apply, i.e., the hosts don't die.

PR #214, opened 2024-02-22
Use explicit cast for rounded numbers

wenzeslaus:
* Replace round by lround to create integer.
* Add static cast to integer for results of existing calls of ceil.
* Explicitly convert to bool using static cast in if-conditions.

These satisfy the -Wfloat-conversion GCC warning for these lines.

### DIFF
--- a/include/pops/actions.hpp
+++ b/include/pops/actions.hpp
@@ -90,7 +90,7 @@ public:
                     // From all the generated dispersers, some go to the soil in the
                     // same cell and don't participate in the kernel-driven dispersal.
                     auto dispersers_to_soil =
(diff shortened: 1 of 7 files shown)


wenzeslaus:
I ended up putting all the changes into one PR which made it easier to test and check the warning. I still may split that for an easier review, but the different types of changes are in separate commits.

wenzeslaus:
This is now ready for review.

I split the PR into multiple PRs and then cleaned up this one. It should be easier to review this way. The other PRs are ready for review as well.

ChrisJones687 (approved):

PR #216, opened 2024-03-07
Add float conversion warning to compile options

wenzeslaus:
Fail compilation when a compiler (GCC or Clang) warning is issued. The original motivation is catching -Wfloat-conversion in CI since we invested into fixing it.


wenzeslaus:
The compilation in CI now fails with:

```
 /home/runner/work/pops-core/pops-core/include/pops/actions.hpp:95:42: error: implicit conversion turns floating-point number into integer: 'double' to 'int' [-Werror,-Wfloat-conversion]
```

which is good because it means that the check and CI works. The errors will be fixed by the float-conversion-related PRs which are now open. After these are merged, this can be merged.

PR #217, opened 2024-03-07
Use explicit floor or ceil for forceful int to double conversions

wenzeslaus:
- Uses static_cast with floor for clear truncation cases (without subtraction).
- Use static_cast with floor or ceil depending on the type of treatment based on the original code (in treatments and tests).


wenzeslaus:
This does not change the results.

I still want to do the actual change to rounding in a different PR, but we can use this PR to discuss what change needs to be done or I can just prepare another one applying rounding everywhere and we would go from there.

PR #218, opened 2024-03-13
Round to the nearest integer value

wenzeslaus:
Instead of floor or ceil, always round to the nearest integer value as soon as the floating point calculation is finished.

This changes results in r.pops.spread and presumably in rpops, too.

### DIFF
--- a/include/pops/host_pool.hpp
+++ b/include/pops/host_pool.hpp
@@ -87,7 +87,8 @@ public:
      *
      * If *use_mortality* is set to false, the *mortality_tracker_vector* is not
      * used or otherwise accessed, and any attempts to retrieve mortality information
-     * will provide empty results (but will not fail).
+     * will provide infected hosts as if there is one cohort or empty results
+     * (but will not fail).
      *
      * Host is added to the environment by the constructor. Afterwards, the environment
      * is not modified.
@@ -561,6 +562,7 @@ public:
      * @brief Completely remove any hosts
      *
      * Removes hosts completely (as opposed to moving them to another pool).
+     * If mortality is not active, the *mortality* parameter is ignored.
      *
      * @param row Row index of the cell
      * @param col Column index of the cell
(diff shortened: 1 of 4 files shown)


wenzeslaus:
I computed a raster which is a difference `original_dead - new_dead`.

There are only cells with -1 in the first year (sum -414):

Screenshot from 2024-03-13 16-43-41

For the last year, it is more variable (min: -10, max: 13, sum: 5650, mean: 0.201901, mean excluding zeros: 1.13614):

Screenshot from 2024-03-13 16-44-42

The pattern copies the data.

The overall impact on the average infected count (5 runs) is significant:

## Before (max 6)

Screenshot from 2024-03-13 17-16-02

## After (max 3.8)

Screenshot from 2024-03-13 17-15-52

wenzeslaus:
## Modified code and modified test

For a selected r.pops.spread test (test_outputs_mortality_treatment, Check mortality together with treatment), the following counts of dead hosts are produced for the original version with floor (and some ceil) and the version with round.

In the test, I changed mortality from 0.5 to 0.4 for the new version with round which creates more aligned totals and the image is visually also more aligned with the original. This similarity is supported by the differences between average of infected.

year | floor, 0.5 | round, 0.4
--- | --- | ---
2019 | 782 | 713
2020 | 2483 | 2374
2021 | 6552 | 6771
2022 | 15037 | 16143

Screenshot from 2024-03-15 10-24-15

### Relative difference for average infected

`old - new) / ((old + new) / 2)`

Screenshot from 2024-03-15 10-06-47

```
n: 15188
minimum: -2
maximum: 2
range: 4
mean: -0.0647932
mean of absolute values: 1.04301
standard deviation: 1.31813
sum: -984.079422039648

total null and non-null cells: 27984
total null cells: 12796
```

### Difference for average infected

`old - new`

Screenshot from 2024-03-15 09-59-26

```
n: 27984
minimum: -4
maximum: 3.2
range: 7.2
mean: -0.0437536
mean of absolute values: 0.272727
standard deviation: 0.512051
sum: -1224.4
```

wenzeslaus:
TODO: Split this into multiple PR to clear identify the source of differences.

wenzeslaus:
This is meant to be closed, but still using it now to track if all planned changes were done in individual PRs.

PR #220, opened 2024-10-01
Allow network node to begin and end edge

wenzeslaus:
This removes the check for a network node being the same at the start and end of a segment (edge). This allows for round trips on a segment which visits some cells but returns to the same cell.

This restriction was already lifted for rpops in https://github.com/ncsu-landscape-dynamics/rpops/commit/3e34ad941e82d14c4797a8a094af5e7489fd62ed


wenzeslaus:
The check was mistakenly added back to rpops in an update of pops-core in https://github.com/ncsu-landscape-dynamics/rpops/commit/ac1383a52b8057091e1dc06e6196c045aeeadfb9.

ChrisJones687 (approved):

PR #224, opened 2024-10-24
Use rounding for mortality

wenzeslaus:
Replaced flooring by rounding in computation of mortality in index/cohort for a cell.


wenzeslaus:
The rounding causes higher mortality which with time causes lower infection counts.

image

image

(infected is an average of number of infected over 5 runs; dead is accumulated number of dead in the first run; 50 runs seed 90385)

wenzeslaus:
The following figures are using mortality rate 0.4 for the new code which brings the results closer to the original code with mortality rate 0.5.

image

image

PR #226, opened 2025-04-29
Add multiple networks

wenzeslaus:
The new MultiNetwork class works with one or more networks. The code structure is changed in a way that network can be switched statically (as a template parameter) instead of using the original network class but having index as its template parameter. This means that model does not know about the number of networks (or what network is actually used). The kernel newly does not know any details about the network and the network itself handles the different movements (which requires the distribution to be instantiated for every call because of constness).

Adds weights for selecting networks, support config in MultiNetwork constructor, a wrapper for random selection by weight and use it in MultiNetwork and in Network, and loading from file for convenience.

Adds tests for one and three networks in MultiNetwork and for the 'pick from list' functions. Adds usage examples to tests to test rpops.

Initialize time-related config variables to avoid warning triggered by a new network test.


ChrisJones687:
- [x] Check with rpops
- [x] check with r.pops.spread

Issue #230, opened 2025-06-12
The size method cannot be used to check for mortality in Rcpp


wenzeslaus:
line 561 of host_pool.hpp
config.use_mortality would fix it

PR #225, opened 2024-10-24
Use rounding for treatments

wenzeslaus:
Replaced flooring by rounding in all treatments.

- [x] PR in rpops with this change https://github.com/ncsu-landscape-dynamics/rpops/pull/203
- [ ] PR in r.pops.spread with this change https://github.com/ncsu-landscape-dynamics/r.pops.spread/pull/79

PR #231, opened 2025-06-13
Explicitly disable mortality in host pool

wenzeslaus:
Instead of relying on size of the mortality vector, ignore the provided data completely when mortality is explicitly disabled. This povides greater flexibility on what rpops and Rcpp can do (#230).

Besides adding yet another config value to the HostPool, add also a constructor which takes a Config object, so that all values can be taken from there. Switch to that in Model and tests when HostPool is used directly, but keep using the old, but updated constructor in tests which don't use config and in Simulation because the Simulation and its tests are not using Config.

### DIFF
--- a/include/pops/config.hpp
+++ b/include/pops/config.hpp
@@ -26,6 +26,7 @@
 
 #include "scheduling.hpp"
 #include "utils.hpp"
+#include "model_type.hpp"
 
 #include <cmath>
 #include <vector>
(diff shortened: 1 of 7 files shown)


ChrisJones687:
Once this is done I can review it, and then can we merge it with PR #225 as I need both at once in order to test if this indeed fixes the issue of treatments not applying with those changes.

ChrisJones687 (approved):
This all looks good. Once merged we will need to make the same change in PR #225 current line number 569.

